Issue #26, opened 2023-01-20
Test API link with an integration test


rmanaem:
The test should only check that the request goes out, not if there is a response.
The code snippet sending the request:
https://github.com/neurobagel/query-tool/blob/156902e5105db12f31032f314d05429f92abb004/components/QueryForm.vue#L178-L184

PR #27, opened 2023-01-22
Implemented UI elements for age

### LINKED ISSUE #14: Implement UI elements for age
- [x] Implement the UI element as a vue component
- [x] Display the component in the QueryFrom.vue component
- [x] Connect the element to an (asynchronous) API request
- [x] Test the implementation with a component test
- [x] Add a short line to the README about what the UI element queries (like an early draft of user docs)

rmanaem:
Closes #14

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,10 +21,11 @@ module.exports = {
   rules: {},
   overrides: [
     {
-      files: ['cypress.config.js', './cypress/support/component.js'],
+      files: './cypress/support/component.js',
       rules: {
         'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
       },
     },
   ],
+  ignorePatterns: 'cypress.config.js',
 };
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@
 Query Tool is a Vue application, developed in [JavaScript](https://www.javascript.com/) using a variety of tools including [Nuxt](https://nuxtjs.org/), [Cypress](https://www.cypress.io/), and [BootstrapVue](https://bootstrap-vue.org/docs).
 
 [Getting started](#getting-started) |
+[Usage](#usage) |
 [Testing](#testing) |
 [License](#license)
 
@@ -71,6 +72,14 @@ npm run build && npm run start
 ```
 
 You can verify the tool is running once you receive info messages from Nuxt regarding environment, rendering, and what port the tool is running on in your terminal.
+
+## Usage
+
+To define a cohort, set your inclusion criteria using the following query fields:
+
+- Age: The minimum and/or maximum age (in years) of participant that should be included in the results.
+
+Once you've defind your criteria, submit them as a query and query tool will display the results.
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -45,6 +45,17 @@ You can do so by running the following command:
 ```bash
 npm install
 ```
+
+### `Set the Environment Variables`
+
+You'll need to set the environment variables (API_QUERY_URL) required to run the query tool. Query tool utilizes [nuxt dotenv module](https://github.com/nuxt-community/dotenv-module) for managing environment variables. 
+
+To set environment variables, create an `.env` file in the root directory and add the environment variables there. If you're running the API locally on your machine, your `.env` file would look something like this:
+
+```bash
+API_QUERY_URL=http://localhost:8000/query/?
+```
+
 ### `Launch Query Tool`
 
 To launch the tool in developer mode run the following command:
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ To define a cohort, set your inclusion criteria using the following query fields
 
 - Age: The minimum and/or maximum age (in years) of participant that should be included in the results.
 
-Once you've defind your criteria, submit them as a query and query tool will display the results.
+Once you've defined your criteria, submit them as a query and query tool will display the results.
 
 ## Testing
 
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -9,30 +9,9 @@
       </b-row>
       <b-row>
         <b-form>
-          <b-form-row class="row">
-            <b-form-group class="col-md-6">
-              <div class="mb-3">
-                <label class="form-label">Age lower</label>
-                <b-form-input
-                  type="number"
-                  name="age_lower"
-                  step="any"
-                  class="numberinput form-control"
-                />
-              </div>
-            </b-form-group>
-            <b-form-group class="col-md-6">
-              <div class="mb-3">
-                <label class="form-label">Age upper</label>
-                <b-form-input
-                  type="number"
-                  name="age_upper"
-                  step="any"
-                  class="numberinput form-control"
-                />
-              </div>
-            </b-form-group>
-          </b-form-row>
+          <age-field
+            @update-age="updateAge"
+          />
           <b-button
             variant="primary"
             type="submit"
@@ -62,5 +41,17 @@
 
 <script>
 export default {
+  data() {
+    return {
+      minAge: null,
+      maxAge: null,
+    };
+  },
+  methods: {
+    updateAge(minAge, maxAge) {
+      this.minAge = minAge;
+      this.maxAge = maxAge;
+    },
+  },
 };
 </script>
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -28,13 +28,19 @@
       <b-row>
         <h2>Results</h2>
       </b-row>
-      <div class="row d-flex justify-content-center">
+      <b-row class="d-flex justify-content-center">
         <div class="d-flex align-items-center">
-          <h3 style="color: gray">
+          <h3
+            v-if="results.length === 0"
+            style="color: gray"
+          >
             Click 'Query Metadata' for results
           </h3>
+          <div v-else>
+            {{ results }}
+          </div>
         </div>
-      </div>
+      </b-row>
     </b-col>
   </b-row>
 </template>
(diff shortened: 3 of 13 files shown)


components/QueryForm.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <b-row class="mx-4">
+    <b-col
+      id="query-fields-column"
```
surchs, line 4: Could you explain why you are setting `id` manually here?

components/QueryForm.vue
```diff
@@ -0,0 +1,83 @@
...
+            Query Metadata
+          </b-button>
+        </b-form>
+      </b-row>
+    </b-col>
+    <b-col
+      id="query-results-column"
+      cols="10"
+    >
+      <b-row>
```
surchs, line 30: I think this would be more maintainable if you turned the results section into its own component and then reference it on the index.vue page. This would also make the current name of this component (`QueryForm`) more specific, because the component would only be about the query form and the logic related to the query there.

components/QueryForm.vue
```diff
@@ -0,0 +1,83 @@
...
+    async submitQuery() {
+      let url = process.env.API_QUERY_URL;
+      if (this.minAge) {
+        url += `min_age=${this.minAge}`;
+      }
+      if (this.maxAge) {
+        url += `&max_age=${this.maxAge}`;
+      }
+      try {
+        const resp = await this.$axios.get(url);
```
surchs, line 73: very neat

cypress/support/bootstrap-vue.js
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue';
```
surchs, line 1: Is this file only necessary because you're using bootstrap as a nuxt-module?

cypress/support/component.js
```diff
@@ -16,6 +16,9 @@
 // Import commands.js using ES2015 syntax:
 import './commands';
 
+// Import BootstrapVue
```
surchs: Can you explain why you need to import bootstrap here? And also why you are importing it from this file (that you created for that exact purpose, I believe)?

cypress/component/AgeField.cy.js
```diff
@@ -0,0 +1,9 @@
+import AgeField from '../../components/AgeField.vue';
+
+describe('Age field', () => {
+  it('Displays two input fields, one for min age and one for max age', () => {
+    cy.mount(AgeField);
+    cy.get("[data-cy='minAge']").should('be.visible');
+    cy.get("[data-cy='maxAge']").should('be.visible');
```
surchs, line 7: visible is indeed useful. But I think it would also make sense to check the functionality you have defined (not the component itself). So e.g.: 
- you want the component to only accept numbers, does it?
- you don't want the value to be negative, what happens when you try anyway?

Ideally you write these tests as you are developing the app, so you can clarify your intention, and then see that the code you write addresses the goal you had set for yourself.

surchs: Oh, and most importantly: you want to make sure that your own component actually emits the correct event with the correct payload in the right situation. You can use the snippets here for some examples: https://github.com/neurobagel/documentation/wiki/Cypress-Test-Snippets#listen-for-an-emitted-event

cypress/component/QueryForm.cy.js
```diff
@@ -0,0 +1,16 @@
...
+};
+
+describe('Query form', () => {
+  it('Displays the AgeField component and submitQuery button', () => {
+    cy.mount(QueryForm, {
+      stubs,
+    });
+    cy.get("[data-cy='ageField']").should('be.visible');
+    cy.get("[data-cy='submitQuery']").should('be.visible');
+  });
```
surchs, line 15: Same thing here, you do have a good bit of functionality now that this component handles. Most importantly the API call. So there should be a test to assert that when I hit the "Query" button, an API call indeed will go out.

.eslintrc.js
```diff
@@ -21,10 +21,11 @@ module.exports = {
...
   overrides: [
     {
-      files: ['cypress.config.js', './cypress/support/component.js'],
+      files: './cypress/support/component.js',
       rules: {
         'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
       },
     },
   ],
+  ignorePatterns: 'cypress.config.js',
```
surchs, line 30: Cool, didn't know about this. Do you prefer that over a dedicated `.eslintignore` file?

README.md
```diff
@@ -61,6 +73,14 @@ npm run build && npm run start
 
 You can verify the tool is running once you receive info messages from Nuxt regarding environment, rendering, and what port the tool is running on in your terminal.
 
+## Usage
```
surchs, line 76: Nice

README.md
```diff
@@ -61,6 +73,14 @@ npm run build && npm run start
...
+
+To define a cohort, set your inclusion criteria using the following query fields:
+
+- Age: The minimum and/or maximum age (in years) of participant that should be included in the results.
+
+Once you've defind your criteria, submit them as a query and query tool will display the results.
```
surchs:
```suggestion
Once you've defined your criteria, submit them as a query and query tool will display the results.
```

surchs (changes requested):
Very cool PR @rmanaem :tada: ! I gave my feedback based on the assumption that we aren't rushing this out - so I commented on your tests and the app structure. Neither of these impacts function. But splitting the components will make the app structure more maintainable, and writing tests that cover the actual functionality will help with development in the future. Ideally the tests are how you specify for yourself what the app will do before you implement that functionality - that's where we want to get.

That said: all looks very nice and functional. Great job!

cypress/support/component.js
surchs: Just to clarify: I mean "can you explain that in this comment", because right now the comment only tells me things I can understand from reading the code itself

components/QueryForm.vue
rmanaem: The `id` is for the custom styles that will be added to the `index` page
I plan to go over them after everything is implemented and remove whatever is not needed.

components/QueryForm.vue
rmanaem: Yea already done, will show up in the next PR.
Just wanted it to look like the old UI before the `Results` component was added.

components/QueryForm.vue
rmanaem: Thanks!

cypress/support/bootstrap-vue.js
rmanaem: No, the file itself is not required but the imports are required (regardless of whether bootstrap is used as a nuxt module or a plugin) for cypress to apply bootstrap styles in tests.

README.md
rmanaem: Good catch!

.eslintrc.js
rmanaem: Yea I prefer to have it all in one place.

cypress/support/component.js
rmanaem: > Can you explain why you need to import bootstrap here? And also why you are importing it from this file (that you created for that exact purpose, I believe)?

For cypress to apply bootstrap styles, bootstrap needs to be imported into cypress (in this case component context). I added the imports as a file to keep `component.js` clean.

> Just to clarify: I mean "can you explain that in this comment", because right now the comment only tells me things I can understand from reading the code itself

Will do

cypress/component/QueryForm.cy.js
rmanaem: >  Most importantly the API call. So there should be a test to assert that when I hit the "Query" button, an API call indeed will go out.

Wouldn't testing that require an integration test?

surchs: Hhm, good question. I don't think so here. You're still just checking the code within the component. I'm thinking about it from the perspective of writing the test before you do the implementation - which is the goal. 

So one would probably be: "when I click the submit button, axios should fire a request somewhere". Doesn't matter if that request is successful, but you want to make sure that your component correctly handles its own responsibilities - which currently include firing axios requests.

I think the integration test would be to assert that provided your axios request was successful (i.e. mocked), some other component of the app would now update with the results (e.g. the results component). So there the concern isn't: "is my component meeting its own responsibilities" but more "is my component playing nice with others"

cypress/component/AgeField.cy.js
rmanaem: I think in the case of AgeField.vue (and perhaps other fields), testing for functionality is trivial since the component is simple. The functionality comes packaged with HTML elements and the component and its implementation doesn't add anything to be tested e.g.,
> you want the component to only accept numbers, does it?

Since the `input` element's `type` property is set to number it doesn't allow for anything other than number input.

> you don't want the value to be negative, what happens when you try anyway?

Something I forgot to take into consideration so thanks for mentioning but it can be added with [`min`](https://www.w3schools.com/tags/att_input_min.asp) property on the `input` element.

That being said I think having an assertion for the emit is a good idea and good practice.

cypress/component/QueryForm.cy.js
rmanaem: > when I click the submit button, axios should fire a request somewhere

I see, do you have any resources on how we could test for that? I've been looking around but haven't been able to find anything on that.

I may have taken your example literally, do you mean test if the `updateAge` function is executed i.e. `spy`?

surchs: Hey, sorry for the late response here. I think the relevant cypress docs would be here: https://docs.cypress.io/guides/guides/network-requests. And if you wanted to also test that the results get rendered after a successful request, you'd need to stub the request (https://docs.cypress.io/guides/guides/network-requests#Stubbing).

I think for now, just asserting that a request is in fact sent to the expected URL would be enough.

cypress/component/AgeField.cy.js
surchs: I think a good rule of thumb is: if it can breaks because you've done something wrong, then test it. So if you have to set a parameter to make it request only positive numbers, then that's a piece of code you wrote. If you ever feel inclined to remove these constraints because you don't remember why they matter, the test will be there for you and yell at you. In that sense the test is just a code form of your requirements: this component should let me enter numbers, but not letters, and the numbers should be positive. 

You shouldn't test things that are maintained by someone else and are their problem if they break. So you shouldn't test is whether adding and removing the integer constraint for this specific component actually allows and disallows you adding things other than numbers.

cypress/component/QueryForm.cy.js
rmanaem: Refer to #26

cypress/component/AgeField.cy.js
rmanaem: Test for text input is implemented.
See #28 for testing negative input.

cypress/component/AgeField.cy.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+  it('Types text into the age input fields', () => {
```
surchs:
```suggestion
  it('Ignores when text is entered into the age input fields', () => {
```

surchs (approved):
Looks great @rmanaem! I made one suggestion for a test name, with that good to go! :tada:

PR #29, opened 2023-01-26
Implemented results UI and its functionalities

### LINKED ISSUE #10: Display response
Given a query result list received from BagelAPI, we want a component that can display visually some basic information about each result dataset object.

- [x] Default view before results are available that indicates how to run a query
- [x] A view If no results are returned (i.e. list is empty)

When a non-empty query is returned display:
- [x] A select all option
- [x] Summary stats
- [x] Download results button
- [x] File download functionality 

For each dataset in the response display a:
- [x] Card
- [x] Dataset title
- [x] Number of subjects
- [x] Modalities toolbar on cards

Test
- [x] default view
- [x] No result view
- [x] Select all functionality: Whether it selects and unselects all checkboxes and updates `download` correctly
- [x] A card displayed for each dataset returned
- [x] Cards display the correct dataset title, number of subjects, and modalities
- [x] Summary stats
- [x] Download results button functionality sync with checkboxes
- [x] `update-download` event is emitted

rmanaem:
closes #10


.eslintrc.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
         'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
       },
     },
+    {
+      files: '*.vue',
+      rules: {
+        'vue/multi-word-component-names': 'off',
```
surchs, line 32: I think this is one of the more "essential" rules Vue recommends. Could you say why you think it's better to disable?

components/DownloadResultsButton.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <b-col
+    class="d-flex flex-row-reverse"
+    style="margin-top: 1em;"
+  >
+    <b-row
+      v-if="!Object.is(results, null) && results.length !== 0"
```
surchs: Subjective view: this is borderline complex for inline code - I would turn this into a computed property with a readable name like `resultsAreReady` or something like it.

components/DownloadResultsButton.vue
```diff
@@ -0,0 +1,82 @@
...
+    downloads: {
+      type: Array,
+      required: true,
+      default: () => [],
+    },
+  },
+  methods: {
+    generateCSVString() {
+      const data = [];
+      this.results.filter((res) => this.downloads.includes(res.dataset_name)).forEach((res) => {
```
surchs: also subjective: I'd line-break `.forEach`... here to improve readability and make it clearer visually where `filter` ends.

components/DownloadResultsButton.vue
```diff
@@ -0,0 +1,82 @@
...
+        res.subject_file_paths.forEach((path) => {
+          data.push({
+            'dataset ': res.dataset_name,
```
surchs:
```suggestion
            'dataset': res.dataset_name,
```

components/DownloadResultsButton.vue
```diff
@@ -0,0 +1,82 @@
...
+          const val = row[header];
+          return `"${val}"`;
+        });
+        csvRows.push(values.join(','));
+      });
+
+      return csvRows.join('\n');
+    },
+
+    download() {
```
surchs: very cool.

components/DownloadResultsButton.vue
```diff
@@ -0,0 +1,82 @@
...
+      csvRows.push(headers.join(','));
+      data.forEach((row) => {
+        const values = headers.map((header) => {
+          const val = row[header];
+          return `"${val}"`;
+        });
+        csvRows.push(values.join(','));
+      });
+
+      return csvRows.join('\n');
```
surchs, line 61: I find this method a bit tricky to read through because the `data` object seems to be created mainly to define a fixed set of headers for the future csv - and to explicitly tie each results attribute to the correct header. I think this is all good, especially if you are doing it to make it easier to add columns cleanly in the future. But it warrants a comment or two - because it reads a bit unexpected (why don't you just build `csvRows` directly like that") and the stuff you need to do to unpack `data` later in the `csvRows` loop is itself a bit tricky to follow.

components/QueryForm.vue
```diff
@@ -1,54 +1,31 @@
...
+          Query Metadata
+        </b-button>
+      </b-form>
+    </b-row>
+  </b-col>
 </template>
 
 <script>
 export default {
+  emits: ['update-results'],
```
surchs, line 28: that's cool. I don't think we've been using this on the annotation_tool.

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+    <b-card>
+      <b-card-body>
+        <b-row>
+          <b-col cols="8">
+            <b-row class="d-flex justify-content-center">
+              <b-col
+                cols="1"
+                class="d-flex align-items-center card-checkbox"
+              >
+                <input
```
surchs, line 12: I'm guessing there is something related to the "select all" functionality that makes it impractical to use the [bootstrap-checkbox component](https://bootstrap-vue.org/docs/components/form-checkbox), right?

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+                  </b-col>
+                </b-row>
+              </b-col>
+            </b-row>
+          </b-col>
+          <b-col cols="4">
+            <b-row>
+              <b-button-toolbar
+                class="d-flex justify-content-end"
+                role="toolbar"
```
surchs: oh, that's interesting. Would you mind explaining what this does? I don't think I've seen this yet.

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+                aria-label="Toolbar with button groups"
+              >
+                <b-button-group
+                  v-for="modal in imageModals"
```
surchs: I think I mentioned this on the API as well, but here it is more applicable: I think you'll prevent some future head-scratching (even short lived ones) if you find a variable name other than modal here because that already [means something else in the context of this codebase](https://bootstrap-vue.org/docs/components/modal).

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+      <b-card-body>
+        <b-row>
+          <b-col cols="8">
+            <b-row class="d-flex justify-content-center">
+              <b-col
+                cols="1"
+                class="d-flex align-items-center card-checkbox"
+              >
+                <input
+                  ref="checkbox"
```
surchs, line 13: When you have many results cards, wouldn't the fixed `ref` here clash as a unique identifier across several instances of this component?

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+                  class="form-check-input dataset-checkbox"
+                  type="checkbox"
+                  value=""
+                  :checked="selectAll"
+                  aria-label="..."
```
surchs: is this a screen reader instruction? If so, then if you use it, I'd say use it with an actual instruction.

components/ResultCard.vue
```diff
@@ -0,0 +1,139 @@
...
+          name: 'EEG',
+          style: 'background-color:rgb(128, 1, 1); color: white;',
+          ariaLabel: 'Fifth group',
+        },
+      },
+    };
+  },
+  methods: {
+    updateDownloads() {
+      this.$emit('update-downloads', this.datasetName, this.$refs.checkbox.checked);
```
surchs, line 125: OK, I think I understand why you need to use refs here. It's because `:checked` already binds the `selectAll` prop, yes? Maybe @jarmoza has an idea about this.

components/Results.vue
```diff
@@ -0,0 +1,126 @@
...
+          v-else-if="results.length === 0"
+          style="color: gray;"
+          data-cy="no-results-view"
+        >
+          No results
+        </h3>
+      </div>
+    </b-row>
+    <b-row
+      v-if="!Object.is(results, null) && results.length !== 0"
```
surchs: subjective: I'd use computed properties with explicit names. You can have them depend on each other too to replicate what you do here

components/Results.vue
```diff
@@ -0,0 +1,126 @@
...
+      id="query-controls"
+    >
+      <b-col cols="7">
+        <input
+          v-model="selectAll"
```
surchs: neat

cypress/component/ResultCard.cy.js
```diff
@@ -0,0 +1,33 @@
...
+
+const props = {
+  datasetName: 'cool-dataset',
+  numMatchingSubjects: 4,
+  imageModals: ['http://purl.org/nidash/nidm#T1Weighted', 'http://purl.org/nidash/nidm#T2Weighted'],
+  selectAll: true,
+};
+
+describe('Result card', () => {
+  it('Displays a card with title, number of subjects, modalities, and checkbox', () => {
```
surchs, line 11: very nice!

cypress/component/ResultCard.cy.js
```diff
@@ -0,0 +1,33 @@
...
+    cy.mount(ResultCard, {
+      listeners: {
+        'update-downloads': cy.spy().as('spy'),
+      },
+      propsData: props,
+    });
+    cy.get("[data-cy='card-cool-dataset-checkbox']").uncheck();
+    cy.get('@spy').should('have.been.calledWith', 'cool-dataset', false);
+    cy.get("[data-cy='card-cool-dataset-checkbox']").check();
+    cy.get('@spy').should('have.been.calledWith', 'cool-dataset', true);
```
surchs, line 31: very nice as well!

cypress/component/Results.cy.js
```diff
@@ -0,0 +1,83 @@
...
+      },
+    });
+    cy.get("[data-cy='no-results-view']").should('be.visible');
+  });
+  it('Checks and unchecks select all checkbox', () => {
+    cy.mount(Results, {
+      stubs,
+      propsData: props,
+    });
+    cy.get("[data-cy='select-all']").should('be.visible').check().should('be.checked');
```
surchs, line 48: oh cool, wasn't aware of check().

pages/index.vue
```diff
@@ -6,12 +6,103 @@
...
+    <b-row class="mx-4">
+      <query-form @update-results="updateResults" />
+      <results :results="results" />
+    </b-row>
   </b-container>
 </template>
 
 <script>
 export default {
-  name: 'IndexPage',
```
surchs, line 15: a name is good, no?

surchs (approved):
Thanks for the PR @rmanaem. This all looks really good. I left a couple of comments that are more suggestions where I think the code could need a comment or a small change to make more readable. 

And: very nice tests!

.eslintrc.js
rmanaem: I disabled it because of `Results.vue` component and `index.vue` page. It felt somewhat trivial to me that component names should be multi-word. I don't feel too strongly about `Results.vue` name but I think `index.vue` shouldn't be forced to be multi-word. what do you think?

components/DownloadResultsButton.vue
rmanaem: Yea that's what I had in mind. Will do.

.eslintrc.js
surchs: Well it's more: I trust the vue-people to know why they add it (https://eslint.vuejs.org/rules/multi-word-component-names.html). I don't think it'll cause any problems, but unless there's a good reason to deviate from their recommendations, why take any chances

components/DownloadResultsButton.vue
rmanaem: I had it that way but eslint flags it as `unnecessarily quoted property` and replaces it with `dataset` property which is not present in the component

pages/index.vue
rmanaem: I guess but other components didn't have it and eslint vue config didn't complain so I decided to remove it.

components/DownloadResultsButton.vue
surchs: Interesting. Does that mean eslint is flagging things it shouldn't? For this specific case: feel free to keep it for now, as far as I understand all this does it make the CSV have a header with an extra space, right?

components/ResultCard.vue
rmanaem: I don't think there is, I'll swap them with their bootstrap alternatives.

components/ResultCard.vue
rmanaem: Good call!

components/ResultCard.vue
rmanaem: I don't think it would since it's handled internally by the component and each card is not aware of the others.

components/ResultCard.vue
rmanaem: The `role` attribute?

surchs: Yeah, I was wondering what consequence it has here.

components/ResultCard.vue
rmanaem: Some of the attributes (including this) and their values were taken from the mock that I did which was based on the concept query code.
I did a quick refactor and tried to remove anything I thought was unnecessary but seem to have missed a few. 
I'm not sure if having this attribute is necessary so I'll just remove all `aria` attributes.

components/ResultCard.vue
rmanaem: Yea `:checked` binds `selectAll` to overwrite manually checking a checkbox but checking/unchecking a checkbox needs to be emitted for each card.

components/ResultCard.vue
rmanaem: It's an attribute incorporated as part of the WAI that I forgot to remove 😄.
It describes the role of an element in (I think only a few) programs like screen readers and magnifiers that can use it.

components/DownloadResultsButton.vue
rmanaem: I think so but I refactored the whole method and it's no longer an issue.

.eslintrc.js
rmanaem: Fair enough, what do you think about disabling it for `index.vue` only (for now)?

components/ResultCard.vue
surchs: OK, let's keep it as is then.

.eslintrc.js
surchs: Ooh, my bad. I missed that. Yeah, of course! Nuxt requires there to be an index.vue component, correct? Then let's disable it for index.vue definitely. 

That reminds me: I am not actually sure why it has to be called index.vue - but I can't find anything that links to it, so that name is probably configured somewhere deep.

components/ResultCard.vue
rmanaem: Well as it turns out simply swapping it won't do. The bootstrap checkbox component breaks functionality.

.eslintrc.js
rmanaem: I don't think nuxt requires it to be called `index.vue`, just need to update nuxt router.
I was saying that for the convention in web dev that the main/home page is usually named `index`

components/ResultCard.vue
rmanaem: I'll keep the current implementation (with `input` element) for now and we can discuss it later.

PR #31, opened 2023-01-31
Implemented UI elements for sex

### LINKED ISSUE #15: Implement UI elements for sex
- [x] Provide options for sex
- [x] Be hooked up to an (asynchronous) API request
- [x] Have a short line added to the README about what the UI element queries (like an early draft of user docs)
- [x] Component test

rmanaem:
Closes #15


plugins/categorical-options.js
```diff
@@ -0,0 +1,17 @@
+import Vue from 'vue';
```
surchs, line 1: Aha. I think you have probably crossed the line to "we can do this without a store".

Let's do this: you can use this now if you also make an issue to refactor this stuff into a proper store and drop the custom plugin / mixin in a future PR.

surchs: For a reference as to why we decided to avoid mixins in the future: https://vuejs.org/guide/reusability/composables.html#vs-mixins

in short: breaks brain during debug, better solutions now exist

rmanaem: We'll address it in #32

PR #35, opened 2023-02-02
Implemented UI element for assessment tool

### LINKED ISSUE #18: Implement UI elements for assessment tool availability
- [x] Add field options to `categoricalOptions` in `index.vue`
- [x] Implement the UI using `CategoricalField.vue` component
- [x] Update `updateCategoricalField` in `QueryForm.vue`
- [x] Update `submitQuery` function in `QueryForm.vue`
- [x] Have a short line added to the README about what the UI element queries
- [x] Update `QueryForm.cy.js`

rmanaem:
Closes #18


components/CategoricalField.vue
```diff
@@ -19,17 +19,14 @@ export default {
     name: {
       type: String,
       default: null,
-      require: false,
```
surchs, line 22: `require: false` is the default, right? Assuming you get rid of this for that reason?

components/QueryForm.vue
```diff
@@ -39,6 +46,12 @@
 
 <script>
 export default {
+  props: {
+    categoricalOptions: {
+      type: Object,
+      default: null,
```
surchs, line 52: Shouldn't this be required? If I don't give you this prop, can your component function at all?

cypress/component/QueryForm.cy.js
```diff
@@ -7,13 +7,36 @@ const stubs = {
   'categorical-field': CategoricalField,
 };
 
+const props = {
```
surchs, line 10: I like handling these as props - much more explicit to think about (and we'll replace it with a store soon enough). 

Related to another comment: if you really want to make the `categoricalOptions` prop optional, then you should have a test here where you do not pass this prop and then check that your component still behaves the way you expect it to in such a situation

pages/index.vue
```diff
@@ -18,6 +21,24 @@ export default {
   data() {
     return {
       results: null,
+      categoricalOptions: {
```
surchs, line 24: :tada: Nice!

surchs (approved):
Looks great @rmanaem, agree with the move mixin -> prop! I made a comment about this new prop being optional, take a look and see if that makes sense to you.

Otherwise good to go! :tada:

components/CategoricalField.vue
rmanaem: Yea [all props are optional by default](https://vuejs.org/guide/components/props.html#prop-validation)

components/QueryForm.vue
rmanaem: Technically it should be (and that's the case for most of the props) but once a prop is required, its default cannot be null anymore.

cypress/component/QueryForm.cy.js
rmanaem: Yea but unfortunately the test breaks without passing the props.

PR #38, opened 2023-02-02
Implemented generic component for continuous fields

### LINKED ISSUE #36: Implement generic component for continuous fields
Implement `ContinuousField.vue` to be used as a generic component for continuous fields similar to `CategoricalField.vue`
`ContinuousField.vue` will replace `AgeFields.vue` component and will be used to implement minimum number of sessions field (#19)

rmanaem:
Closes #36


components/ContinuousField.vue
```diff
@@ -0,0 +1,43 @@
...
+    />
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    name: {
+      type: String,
+      default: null,
```
surchs: `null` isn't a good default name. This should either be required or have a sensible default

components/ContinuousField.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <div class="mb-3">
+    <label
+      data-cy="continuous-field-label"
+      class="form-label"
+    >{{ name }}</label>
+    <b-form-input
+      v-model="input"
```
surchs: if you need to handle the `@input` event with a specific handler method, then `v-model` is not the way to go. Use `"value` and `@input` instead. This should also obviate the need for the `input: null` data property

components/ContinuousField.vue
```diff
@@ -0,0 +1,43 @@
...
+<script>
+export default {
+  props: {
+    name: {
+      type: String,
+      default: null,
+    },
+    step: {
+      type: String,
+      default: null,
```
surchs: should be something sensible (e.g. `1`?)

components/ContinuousField.vue
```diff
@@ -0,0 +1,43 @@
...
+    },
+  },
+  emits: ['update-continuous-field'],
+  data() {
+    return {
+      input: null,
```
surchs: why is this `null`? If I understand the use above, this shouldn't be necessary when you replace `v-model` with binding `value` and `@input` separately (as you should)

components/ContinuousField.vue
```diff
@@ -0,0 +1,43 @@
...
+    };
+  },
+  methods: {
+    updateField() {
+      this.$emit('update-continuous-field', this.name, this.input);
```
surchs: instead of `this.input` you can use the content of the `$event` payload from the `@input` event https://bootstrap-vue.org/docs/components/form-input#native-and-custom-events

components/QueryForm.vue
```diff
@@ -64,20 +78,22 @@ export default {
     };
   },
   methods: {
-    updateAge(minAge, maxAge) {
-      this.minAge = minAge;
-      this.maxAge = maxAge;
-    },
-    updateCategoricalField(name, selected) {
+    updateField(name, input) {
```
surchs, line 81: OK, but I think we're reaching the end of how much we should handle with this generic event handler. Let's revisit after the milestone

cypress/component/ContinuousField.cy.js
```diff
@@ -0,0 +1,27 @@
+import ContinuousField from '../../components/ContinuousField.vue';
```
surchs, line 1: Not checking this in detail under the assumption it's identical to the deleted `AgeFields` test. Is that correct?

surchs (changes requested):
Looks good, thanks @rmanaem. But have some change requests (mainly relating to defaults and the event handling on the input field in the continuous compoent. Take a look at lmk what you think

cypress/component/ContinuousField.cy.js
rmanaem: Yea

components/QueryForm.vue
rmanaem: See #39

components/ContinuousField.vue
```diff
@@ -21,22 +20,17 @@ export default {
   props: {
     name: {
       type: String,
-      default: null,
+      required: true,
     },
     step: {
       type: String,
-      default: null,
+      default: '1',
```
surchs, line 27: does this expect a string or a number?

components/ContinuousField.vue
```diff
@@ -21,22 +20,17 @@ export default {
...
-  data() {
-    return {
-      input: null,
-    };
-  },
   methods: {
-    updateField() {
-      this.$emit('update-continuous-field', this.name, this.input);
+    updateField(value) {
+      this.$emit('update-continuous-field', this.name, value);
```
surchs, line 33: neat!

surchs (approved):
Looks great!

One comment on the type of the prop for continuousfield (more of a sanity check) and then good to go :tada:

components/ContinuousField.vue
rmanaem: Yea

PR #44, opened 2023-02-10
Implemented query form validation

### LINKED ISSUE #28: Validate query form
Input fields in the query form need to be validated before the form is submitted.

Cases to keep in mind for validation:
- Age
  - [x] `minAge` < `maxAge`
  - [x] ~both `minAge` and `maxAge` must be non-negative~
    ~currently, `min` property of the `input` element enforces this but this can't be tested inside the AgeField.cy.js test~

Should the validations fail, the user shouldn't be able to submit the form, Perhaps the button becomes `disabled` and the field with the issue is highlighted and an informative message is given to the user.

Implementation
- [x] Implement `displayToast` method to be called from inside the `validateQueryForm`
  - display an informative message using [bootstapvue toast](https://bootstrap-vue.org/docs/components/toast)
- [x] Implement `validateQueryForm` method
  - To be called when the `submitQuery` button Is clicked
  - Call `displayToast` method if there is an error otherwise call  `submitQuery` method
- [x] Update `QueryForm.cy.js` to test the above functionalities

rmanaem:
Closes #28

### DIFF
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -56,4 +56,15 @@ describe('Query form', () => {
     cy.get('[data-cy="healthy-control-checkbox"]').check();
     cy.get('[data-cy="Diagnosis-select"]').should('have.class', 'vs--disabled');
   });
+  it('Calls displayToast method', () => {
+    cy.spy(QueryForm.methods, 'displayToast').as('spy');
+    cy.mount(QueryForm, {
+      stubs,
+      propsData: props,
+    });
+    cy.get('[data-cy="Min Age-continuous-field-input"]').type(7);
+    cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
+    cy.get('[data-cy="submit-query"]').click();
+    cy.get('@spy').should('have.been.calledWith', 'Maximum age must be greater than or equal to minimum age');
+  });
 });
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -66,5 +66,7 @@ describe('Query form', () => {
     cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
     cy.get('[data-cy="submit-query"]').click();
     cy.get('@spy').should('have.been.calledWith', 'Maximum age must be greater than or equal to minimum age');
+    // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
+    cy.contains('#b-toaster-top-right', 'Maximum age must be greater than or equal to minimum age');
   });
 });
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -57,7 +57,6 @@ describe('Query form', () => {
     cy.get('[data-cy="Diagnosis-select"]').should('have.class', 'vs--disabled');
   });
   it('Calls displayToast method', () => {
-    cy.spy(QueryForm.methods, 'displayToast').as('spy');
     cy.mount(QueryForm, {
       stubs,
       propsData: props,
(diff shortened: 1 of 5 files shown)


cypress/component/QueryForm.cy.js
```diff
@@ -56,4 +56,15 @@ describe('Query form', () => {
...
+  it('Calls displayToast method', () => {
+    cy.spy(QueryForm.methods, 'displayToast').as('spy');
+    cy.mount(QueryForm, {
+      stubs,
+      propsData: props,
+    });
+    cy.get('[data-cy="Min Age-continuous-field-input"]').type(7);
+    cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
+    cy.get('[data-cy="submit-query"]').click();
+    cy.get('@spy').should('have.been.calledWith', 'Maximum age must be greater than or equal to minimum age');
```
surchs: This is very cool! But Because your UI interaction has a UI result, I think you should mainly test that the toast is actually shown in the way you expect (something says "error", and the message has some relevant text). The fact that you know what method has to be called internally by your component to make this happen is an implementation detail that you shouldn't have to worry / know about in a component test. So:

- yes: assert that something does in fact pop up on the users screen when you click that button
- no: couple the test closely to the (very beautiful) implementation you have chosen for the component (and happen to aware of because you have already done that work)

In fact, as cool as it is that we can spy component methods like this (didn't know that), I think we should generally not do that for the above reasons.

surchs (changes requested):
Thanks @rmanaem for this really cool PR! This looks and works great. I even learned some new things (you can spy component method calls?). 

I left one comment to test UI, not implementation in the query-form test. Functionality wise all is good, but the test should be done with the eyes of a user (who most likely doesn't know about all the elegant things going on behind the scenes) and just wants to know what pops up and changes outside of the component.

Take a look, and with that change: good to go!

cypress/component/QueryForm.cy.js
rmanaem: Yea I agree and that was the plan initially but since the toast is made dynamically and only lasts a few seconds it's hard to `cy.get()` it and assert on it. So I went with the next best thing, assume that the `displayToast` works as promised and test if it's called when validation fails.

rmanaem: The workaround here is to get rid of `displayToast` and integrate `b-toast` element with the same functionality.

surchs: Oh, interesting. Do you think the toast is too fast for cypress, or rather too slow (e.g. https://stackoverflow.com/a/71301155/1302009)? Too fast would be disappointing (from cypress). Too slow might make sense (see stackoverflow). I'm not sure why using the `b-toast` component would fix the problem.

OK, I'd say:
1. See (briefly) if there is a way for cypress to catch the toast via the UI
2. if there isn't, go with your workaround but make sure to comment it. As you said yourself, you had to deviate from the expected pattern, and that's usually a good sign that you need to comment why you're doing it in an unexpected way

rmanaem: Turns out `b-toast` [isn't the fix I thought it'd be](https://bootstrap-vue.org/docs/components/toast#b-toast-component) but

> https://stackoverflow.com/a/71301155/1302009

this did it.

surchs: very nice! I'd still add a comment to explain why you end up using `cy.contains` over the usual `cy.get`.

rmanaem: Yea included the link to the StackOverflow issue

cypress/component/QueryForm.cy.js
```diff
@@ -56,4 +56,17 @@ describe('Query form', () => {
...
```
surchs: OK, looks great. But now we can remove the component method spy, right? The main point here is: this test shouldn't break just because you rename the method inside of the component that fires the toast.

rmanaem: Yea but I kinda want to keep it as a use-case demonstration of spying on methods.

surchs: I understand. But the right place for that is in the docs. For now that'd be here: https://github.com/neurobagel/documentation/wiki/Cypress-Test-Snippets. So let's add the snippet there (with a little note to use carefully) and then remove it from this test and merge

surchs (approved):
:tada:, very cool PR. Good to go

PR #46, opened 2023-02-10
Handled possible server-side errors

### LINKED ISSUE #25: Handling possible server-side errors
Given the implementation `submitQuery` function in `QueryForm.vue` component, how should server-side errors thrown by our API e.g., [`min_age > max_age` ](https://github.com/neurobagel/bagelapi/blob/fc48a2c2b53bd1907b9825505024a93b9f6ec4f3/app/api/models.py#L22-#L39) be handled?

rmanaem:
Closes #25

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -92,7 +92,7 @@ export default {
       default: null,
     },
   },
-  emits: ['update-results'],
+  emits: ['update-response'],
   data() {
     return {
       minAge: null,
@@ -180,11 +180,10 @@ export default {
       try {
         const resp = await this.$axios.get(url);
         this.results = resp.data;
-        this.$emit('update-results', this.results);
+        this.$emit('update-response', this.results, null);
       } catch (err) {
-        // TODO Handle situations that may throw an error e.g., when min age > max age
-        // could emit an event here and show an "error view" in results or use vuelidate
-        // to validate the form before submission
+        this.results = null;
+        this.$emit('update-response', this.results, 'Oops, something went wrong');
       }
     },
   },
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -11,7 +11,14 @@
     >
       <div class="d-flex align-items-center">
         <h3
-          v-if="Object.is(results, null)"
+          v-if="!Object.is(error, null)"
+          style="color: gray"
+          data-cy="error-view"
+        >
+          {{ error }}
+        </h3>
+        <h3
+          v-else-if="Object.is(results, null)"
           style="color: gray"
           data-cy="default-view"
         >
@@ -85,6 +92,10 @@ export default {
   props: {
     results: {
       type: Array,
+      default: null,
+    },
+    error: {
+      type: String,
       default: null,
     },
   },
--- a/cypress/component/ResultsContainer.cy.js
+++ b/cypress/component/ResultsContainer.cy.js
@@ -28,11 +28,20 @@ const props = {
 };
 
 describe('Results', () => {
-  it('Displays default view of results component', () => {
+  it('Displays error view of results container component', () => {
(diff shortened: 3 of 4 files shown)


components/QueryForm.vue
```diff
@@ -180,11 +180,10 @@ export default {
       try {
         const resp = await this.$axios.get(url);
         this.results = resp.data;
-        this.$emit('update-results', this.results);
+        this.$emit('update-response', this.results, null);
       } catch (err) {
-        // TODO Handle situations that may throw an error e.g., when min age > max age
-        // could emit an event here and show an "error view" in results or use vuelidate
-        // to validate the form before submission
+        this.results = null;
```
surchs, line 185: I don't think anything else inside of this component uses `this.results`. If I understand correctly, this does:

- call API
- if successfull, pass the `response.data` object to the emitted event as payload
- if unsuccessful, pass `null` to the emitted event as payload

so there is no reason for the component to keep track of the results in its state after they have been passed off to a listening parent via the event. 

Unless there's a reason to keep track of `this.results` here that I don't see, I think it's best if we remove this variable and just pass the `response.data` / `null` payload explicitly.

components/ResultsContainer.vue
```diff
@@ -11,7 +11,14 @@
     >
       <div class="d-flex align-items-center">
         <h3
-          v-if="Object.is(results, null)"
+          v-if="!Object.is(error, null)"
```
surchs, line 14:
```suggestion
          v-if="error !== null"
```
Any reason not to just do this? If I understand this correctly, the only reason to use `Object.is` here is if you would like to use `NaN` as a signal for "no error", because that'd behave differently in `===`. But I don't think you're using `NaN` and if you did, we should probably find another signal for "no error" anyway.

I'd prefer `===` just because more commonly used here, i.e. more expected and readable. LMK if there is a strong reason not to do that.

cypress/component/ResultsContainer.cy.js
```diff
@@ -28,11 +28,20 @@ const props = {
 };
 
 describe('Results', () => {
-  it('Displays default view of results component', () => {
+  it('Displays error view of results container component', () => {
+    cy.mount(ResultsContainer, {
+      propsData: {
+        results: null,
+        error: 'Cool error',
```
surchs, line 35: neat!

cypress/component/ResultsContainer.cy.js
```diff
@@ -28,11 +28,20 @@ const props = {
...
+      },
+    });
+    cy.get('[data-cy="error-view"]').should('be.visible').contains('Cool error');
+  });
+  it('Displays default view of results container component', () => {
     cy.mount(ResultsContainer);
     cy.get('[data-cy="default-view"]').should('be.visible');
   });
-  it('Displays no results view of results component', () => {
+  it('Displays no results view of results container component', () => {
```
surchs:
```suggestion
  it('Displays the no results view if results are empty without error', () => {
```

surchs (approved):
Looks good @rmanaem. I left two suggestions where I think you could make the code clearer. Take a look and see if you agree. 

With that, good to go.

components/QueryForm.vue
rmanaem: > I don't think anything else inside of this component uses this.results

You are right but results gets emitted to `index.vue` and then passed as a prop from `index.vue` to `ResultsContainer.vue` where it's used as a dependency of the `displayResults` computed property.
`results` needs to be updated when an error occurs because in case an error occurs after a successful query was run and results were displayed since there was no change in the `results` the `displayResults` property won't be triggered and the error view won't be displayed.

components/ResultsContainer.vue
rmanaem: There isn't a strong reason (to do or) not to do that, just multiple ways of doing (in this case) the same thing.

PR #47, opened 2023-02-12
Refactored codebase

### LINKED ISSUE #40: Refactor before deployment
- [x] Go through existing props
  - Set them to required if the component doesn't work without them
  - or provide a sensible default for them
- [x] Change `Cohort Definition Tool` to `Query Tool` in `index.vue`
- [x] Rename `Modality` to `Imagin modality`
- [x] Rename `updateCategoricalField` method in `CategoricalField.vue` to `updateField`
- [x] ~Add bagel favicon~
- [x] Update the modality colors
- [x] ~Update README.md to follow the same pattern as the api~
- [x] ~Add loading animation while the result is being retrieved~

rmanaem:
Closes #40

### DIFF
--- a/components/CategoricalField.vue
+++ b/components/CategoricalField.vue
@@ -14,7 +14,7 @@
       :data-cy="`${name}-select`"
       :options="options"
       :disabled="disabled"
-      @input="updateCategoricalField"
+      @input="updateField"
     />
   </div>
 </template>
@@ -46,7 +46,7 @@ export default {
     };
   },
   methods: {
-    updateCategoricalField() {
+    updateField() {
       this.$emit('update-categorical-field', this.name, this.selected);
     },
   },
--- a/components/CategoricalField.vue
+++ b/components/CategoricalField.vue
@@ -24,7 +24,7 @@ export default {
   props: {
     name: {
       type: String,
-      default: null,
+      required: true,
     },
     defaultSelected: {
       type: String,
@@ -32,7 +32,7 @@ export default {
     },
     options: {
       type: Array,
-      default: null,
+      default: Array,
     },
     disabled: {
       type: Boolean,
--- a/components/CategoricalField.vue
+++ b/components/CategoricalField.vue
@@ -32,7 +32,7 @@ export default {
     },
     options: {
       type: Array,
-      default: Array,
+      default: () => [],
     },
     disabled: {
       type: Boolean,
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -32,7 +32,7 @@ export default {
     },
     downloads: {
       type: Array,
-      default: Array,
+      default: () => [],
     },
   },
   computed: {
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -28,11 +28,11 @@ export default {
   props: {
     results: {
       type: Array,
-      default: null,
+      required: true,
     },
     downloads: {
       type: Array,
-      default: null,
+      default: Array,
     },
   },
   computed: {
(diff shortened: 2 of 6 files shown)


components/CategoricalField.vue
```diff
@@ -24,15 +24,15 @@ export default {
...
+      required: true,
     },
     defaultSelected: {
       type: String,
       default: null,
     },
     options: {
       type: Array,
-      default: null,
+      default: Array,
```
surchs:
```suggestion
      default: [],
```
careful, you are referencing the builtin `Array` type constructor here, which will turn `options` into a pointer to the [`Array` constructor function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array), and not initialize an empty array as you intended! 

There are a couple of ways to make [a new Array instance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#constructor) but the correct use here is going to be `default: []` because it is readable, short, and expected.

components/DownloadResultsButton.vue
```diff
@@ -28,11 +28,11 @@ export default {
   props: {
     results: {
       type: Array,
-      default: null,
+      required: true,
     },
     downloads: {
       type: Array,
-      default: null,
+      default: Array,
```
surchs:
```suggestion
      default: [],
```

surchs (changes requested):
Looks good @rmanaem. I made a change suggestion for the prop defaults. Take a look.

components/CategoricalField.vue
rmanaem: Yea that's what I did when I first started implementing components and their props but according to [`vue/require-valid-default-prop`](https://eslint.vuejs.org/rules/require-valid-default-prop.html) the default for `Array` and `Object` must be returned using a function.
I had tried using the Array constructor itself but that triggers the `no-array-constructor` error from eslint so I'm passing a pointer to a function instead and the options were what you see here and
```
() => []
```

surchs: Ah I see, my bad. I had forgotten about the Vue props validation. You are right that for Array types the default value cannot be a simple instance. What's wrong with `() => []`? That looks correct to me.

rmanaem: Nothing just another way of doing the same thing

surchs: Hhm, OK. I see what you mean. I don't think these two approaches are equivalent though in terms of readability. We're not trying to turn this codebase into an intelligence test. (if you, unlike me don't forget how Vue props work) the expected syntax for array defaults is an [anonymous function that returns the default](https://vuejs.org/guide/components/props.html#prop-validation). So we should stick to that.

rmanaem: > the expected syntax for array defaults is an [anonymous function that returns the default](https://vuejs.org/guide/components/props.html#prop-validation). So we should stick to that.

Are you referring to this?
```
    propE: {
      type: Object,
      // Object or array defaults must be returned from
      // a factory function. The function receives the raw
      // props received by the component as the argument.
      default(rawProps) {
        return { message: 'hello' }
      }
    },
```

surchs: yeah, although I think I shared the vue3 docs. Here is the relevant section for us: https://v2.vuejs.org/v2/guide/components-props.html#Prop-Validation

rmanaem: Yea thanks for that I didn't know they mentioned that in their docs.

surchs: NP, neither did I :) or rather forgot again. We do use this pattern for defaults in the annotation_tool.

surchs (approved):
:tada: good to go, Thanks for doing this!

PR #49, opened 2023-02-14
Set up lint staged

### LINKED ISSUE #30: Install and configure lint-staged
Right now eslint doesn't understand the difference between the git staging area and the working directory. The consequence is that if I make a partial commit (i.e. I have more changes in my working dir than I include in the next commit), eslint will not limit itself to the files that are included in the commit.

We should install https://github.com/okonet/lint-staged and configure it so it can run as part of our husky pre-commit hooks so we get more informative eslint errors.

rmanaem:
Closes #30

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run lint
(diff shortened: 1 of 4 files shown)


surchs (approved):
Aha, very cool. Thanks @rmanaem. GTG :tada:

PR #52, opened 2023-02-15
Added lint workflow and fixed cd workflow

### LINKED ISSUE #50: Set up workflow for eslint
Set up a workflow to run eslint on PRs to catch any issues in case eslint wasn't run before committing.

### LINKED ISSUE #51: Fix cd workflow
Every time the workflow is run, the domain is reset back to GitHub pages default and to set it to a specific domain a `CNAME` file including the domain needs to be added to the build.

rmanaem:
Closes #50 
Closes #51

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -11,7 +11,7 @@ jobs:
    runs-on: ubuntu-latest
    steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v3
 
     - name: Set up node env
       uses: actions/setup-node@v2.1.2
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -8,13 +8,10 @@ on:
 
 jobs:
   cd:
-
    runs-on: ubuntu-latest
-
    steps:
-
     - name: Checkout
-      uses: actions/checkout@main
+      uses: actions/checkout@master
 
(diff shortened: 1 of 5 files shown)


.github/workflows/cd.yaml
```diff
@@ -8,13 +8,10 @@ on:
...
 jobs:
   cd:
-
    runs-on: ubuntu-latest
-
    steps:
-
     - name: Checkout
-      uses: actions/checkout@main
+      uses: actions/checkout@master
```
surchs: why are we (if I understand correctly) tracking a branch rather than a version here?

surchs (approved):
Looks good @rmanaem. If left a comment for the `cd` workflow because I didn't understand why we're tracking a branch there. Take a look if there is a reason / that makes sense, and then good to go :tada:.

.github/workflows/cd.yaml
rmanaem: I used the yaml script in [Nuxt docs](https://nuxtjs.org/deployments/github-pages/#github-actions). 
After looking into it I realized its the same thing. master is the same as v2 and main is v3

rmanaem: I'll change it to v3 for consistency's sake

surchs: That's good. Tracking a version is preferable over tracking a branch because branches can change and might introduce changes that break things.

PR #59, opened 2023-02-28
Fixed diagnosis and healthy control bug

### LINKED ISSUE #58: Diagnosis and healthy control bug
To reproduce the bug:
Select a diagnosis other than `All`, check `healthy control` checkbox and you'll have a 422 error which is coming from the API validation, see [here](https://github.com/neurobagel/api/blob/3aa954dbd2c7efb801223cfa5f171b376530046f/app/api/models.py#L41-L48)

rmanaem:
Main change:
- Fixed the bug which Closes #58 
- Changed the `results` prop in DownloadResultsButton.vue to not be required and set its default to empty array
   - To resolve the vue warning regarding `results` prop type check

### DIFF
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -28,7 +28,7 @@ export default {
   props: {
     results: {
       type: Array,
-      required: true,
+      default: () => [],
     },
     downloads: {
       type: Array,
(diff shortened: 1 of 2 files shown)


jarmoza (approved):
👨🏽‍🍳

@rmanaem As I was mentioning, it would be good to have a note about how the .env file should be configured depending on where the API is running - locally or remote.

PR #61, opened 2023-03-07
Updated README.md

### LINKED ISSUE #60: Update README.md
Clarify what the `API_QUERY_URL` env var does and mention that the content of the file should be changed based on whether you want the app to contact a local or remote api

rmanaem:
Clarified what the `API_QUERY_URL` env var does and how the `.env` file should look like when using remote API
Closes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,18 @@ A good way to manage different node versions is to use the [node version manager
 
 ### Set the Environment Variables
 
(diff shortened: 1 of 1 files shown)


jarmoza (changes requested):
Just a note to refer to the query tool in a consistent way. Is it "query tool" or "the query tool"? It looks like the intent is to refer to it as "the query tool." There are (at least) two instances when "the" is missing. Other than that,🧑🏼‍🍳.

jarmoza (approved):
👨‍🍳

PR #63, opened 2023-04-24
Updated dropdown options

### LINKED ISSUE #37: Update dropdown options
Update the options in `CategoricalOptions` in `index.vue` once the data has been added to graph.

### LINKED ISSUE #62: `Sex` query field not working as intended
Currently setting the `sex` query field doesn't update the results. This is due to the recent update in the graph and needs to be addressed first in the API.

rmanaem:
Closes #37 
Closes #62


pages/index.vue
```diff
@@ -28,21 +28,38 @@ export default {
...
+          Migraine: 'snomed:37796009',
+          'Concussion injury of brain': 'snomed:110030002',
+          'Autism spectrum disorder': 'snomed:35919005',
+          Dyslexia: 'snomed:59770006',
+          'Specific spelling disorder': 'snomed:268738002',
+          Schizophrenia: 'snomed:58214004',
+          'Borderline personality disorder': 'snomed:20010003',
         },
         'Assessment tool': {
-          All: null,
```
surchs, line 41: do we need to keep `All: null` here so the thing still shows up as an (empty) dropdown?

surchs (approved):
Looks great @rmanaem. Just a sanity check: I don't need to hardcode / provide a controlled term for the status of "I am a control subject", because you just send this as a boolean query property to the API and the API knows what controlled term to put in the SPARQL query if the user asks for "control subjects", correct?

If so, good to go :cook:

pages/index.vue
rmanaem: No, apparently there is no need to keep anything in there, It renders as an empty dropdown.

rmanaem:
> Just a sanity check: I don't need to hardcode / provide a controlled term for the status of "I am a control subject", because you just send this as a boolean query property to the API and the API knows what controlled term to put in the SPARQL query if the user asks for "control subjects", correct?

Yea, that's correct.

PR #65, opened 2023-04-25
Adding Parkinson as diagnosis

surchs:
No tracking issue, just added Parkinson's diagnosis (because we have those data).

- also alphabetized the diagnosis keys


rmanaem (approved):
👨🏼‍🍳

surchs:
Thanks @rmanaem , I'll wait with meeting this until we have discussed the performance issues. Seems like our Parkinson subjects are linked to slow queries

surchs:
I'm going to close this PR for now. I think we will have to address this as part of the openneuro graph deployment

rmanaem:
@surchs Do we need to keep the branch?

PR #68, opened 2023-05-10
Updated the `generateCSVString` function

### LINKED ISSUE #64: Update `generateCSVString`
Update the `generateCSVString` function to reflect the changes made to the API in [this PR](https://github.com/neurobagel/api/pull/101).

rmanaem:
Closes #64

### DIFF
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -42,12 +42,12 @@ export default {
   },
   methods: {
     generateCSVString() {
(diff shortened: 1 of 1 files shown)


surchs (approved):
Changes look good @rmanaem. Just one question: there is currently no e2e test to ensure that the query tool actually understands a response from the API, right? The one e2e test I can see only asserts that the call goes out correctly.

edit: good to go :cook:

rmanaem:
@surchs Yea, you're right there currently isn't an e2e for that.

PR #91, opened 2023-05-29
[ENH] Modified tool for new dataset-level and subject-level results

### LINKED ISSUE #76: Modify tool for new subject level results
To implement the functionality for specs outlined [here](https://github.com/neurobagel/documentation/wiki/Query-Tool):

- [x] Implement a check box to switch between the two `.csv` files
- [x] Modify `generateCSVString` to return new subject-level results (the two `.csv` files).
- [x] Update `DownloadResultsButton` component test
- [x] Update README.md to mention what the tool now returns based on the status of the checkbox

As part of this, we need to make sure that we return the human-readable labels to the users, not the IRI for diagnosis levels and sex. See https://github.com/neurobagel/project/issues/47

For the moment, we will use the hard-coded mapping of controlled term IRI to human-readable label that already exists in the query tool

rmanaem:
Closes #76
Closes https://github.com/neurobagel/project/issues/118

- Implemented Checkbox to toggle results csv
- Modified `generateCSVString` method to return new dataset-level and subject-level results (the two .csv files).
- Updated `DownloadResultsButton` component test
- Update README.md to mention what the tool now returns based on the status of the checkbox
- Implemented spinner fo the submit query button

## Checklist

- [ ] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,11 +111,11 @@ To define a cohort, set your inclusion criteria using the following:
 
 
 Once you've defined your criteria, submit them as a query and the query tool will display the results.\
-The query tool offers two different CSV files for results:
-- Dataset-level results csv contains: dataset portal uri, dataset file path, dataset name, number of matching subjects, and available imaging modalities
-- Subject-level results csv contains: dataset file path, subject id, age, sex, diagnosis, session id, session file path, number of sessions, and imaging modality
+The query tool offers two different TSV files for results:
+- Dataset-level results TSV contains: dataset portal uri, dataset file path, dataset name, number of matching subjects, and available imaging modalities
+- Participant-level results TSV contains: dataset file path, subject id, age, sex, diagnosis, session id, session file path, number of sessions, and imaging modality
 
-You can choose which file to download by checking and uncehcking the `Toggle Results CSV` checkbox above the `Download Results` button.
+You can choose which file to download by checking and uncehcking the `Toggle Results TSV` checkbox above the `Download Results` button.
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -110,7 +110,12 @@ To define a cohort, set your inclusion criteria using the following:
 - Modality: Imaging modality of participant scans that should be included in the results.
 
 
-Once you've defined your criteria, submit them as a query and the query tool will display the results.
+Once you've defined your criteria, submit them as a query and the query tool will display the results.\
+The query tool offers two different CSV files for results:
+- Dataset-level results csv contains: dataset portal uri, dataset file path, dataset name, number of matching subjects, and available imaging modalities
+- Subject-level results csv contains: dataset file path, subject id, age, sex, diagnosis, session id, session file path, number of sessions, and imaging modality
+
+You can choose which file to download by checking and uncehcking the `Toggle Results CSV` checkbox above the `Download Results` button.
 
 ## Testing
 
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -16,7 +16,7 @@
         >
         <label
           class="form-label"
-        >Toggle Results TSV</label>
+        >Toggle level of detail to download</label>
       </b-col>
 
       <b-button
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -5,8 +5,18 @@
     style="margin-top: 1em;"
   >
     <b-row
-      class="mr-auto"
+      class="flex-column"
     >
+      <b-col>
+        <input
+          class="form-check-input"
+          type="checkbox"
+        >
+        <label
+          class="form-label"
+        >Toggle Results CSV</label>
+      </b-col>
+
       <b-button
         id="download-results-button"
         :disabled="downloads.length === 0"
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -106,7 +106,11 @@ export default {
       const element = document.createElement('a');
       element.setAttribute('href', `data:text/tsv;charset=utf-8,
       ${encodeURIComponent(this.generateTSVString())}`);
-      element.setAttribute('download', 'results.tsv');
+      if (this.toggleResultsTSV) {
+        element.setAttribute('download', 'participant-results.tsv');
+      } else {
+        element.setAttribute('download', 'dataset-results.tsv');
+      }
 
       element.style.display = 'none';
       document.body.appendChild(element);
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -9,14 +9,14 @@
     >
       <b-col>
         <input
-          v-model="toggleCSV"
-          data-cy="toggle-csv-checkbox"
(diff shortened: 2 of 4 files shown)


surchs:
Nice, do you want to link https://github.com/neurobagel/project/issues/118 to close it as well?

components/DownloadResultsButton.vue
```diff
@@ -17,7 +29,7 @@
           icon="download"
           font-scale="1"
         />
-        Download Results
+        {{ toggleDownloadResultsButtonText() }}
```
surchs: neat. interesting solution.

components/DownloadResultsButton.vue
```diff
@@ -35,26 +47,57 @@ export default {
...
   },
   methods: {
+    toggleDownloadResultsButtonText() {
+      return this.toggleCSV ? 'Download Subject-level Results' : 'Download Dataset-level Results';
+    },
     generateCSVString() {
-      const headers = ['dataset', 'number of matching subjects', 'subject data'];
-      const csvRows = [headers];
+      const csvRows = [];
+      const datasets = this.results.filter((res) => this.downloads.includes(res.dataset_name));
```
surchs, line 66: :tada: I see it, I get it, very neat

README.md
```diff
@@ -110,7 +110,12 @@ To define a cohort, set your inclusion criteria using the following:
 - Modality: Imaging modality of participant scans that should be included in the results.
 
 
-Once you've defined your criteria, submit them as a query and the query tool will display the results.
+Once you've defined your criteria, submit them as a query and the query tool will display the results.\
+The query tool offers two different CSV files for results:
+- Dataset-level results csv contains: dataset portal uri, dataset file path, dataset name, number of matching subjects, and available imaging modalities
+- Subject-level results csv contains: dataset file path, subject id, age, sex, diagnosis, session id, session file path, number of sessions, and imaging modality
+
+You can choose which file to download by checking and uncehcking the `Toggle Results CSV` checkbox above the `Download Results` button.
```
surchs: 1) I think this is very cool
2) I think we should find a better name for it. I read the PR description, and then the README, and then the code, but I was still a bit unclear what `toggleCSV` toggles between. .tsv -> .csv?

Maybe dataset-level vs participant-level?

components/DownloadResultsButton.vue
```diff
@@ -35,26 +47,57 @@ export default {
...
+    };
+  },
   computed: {
     displayDownloadResultsButton() {
       return !Object.is(this.results, null) && this.results.length !== 0;
     },
   },
   methods: {
+    toggleDownloadResultsButtonText() {
+      return this.toggleCSV ? 'Download Subject-level Results' : 'Download Dataset-level Results';
```
surchs: Ah OK, I guess in the interface this will be a lot clearer. Cool. Then I think: let's also make it clearer inside the codebase. Maybe rename: `this.toggleParticipantResults`?

components/DownloadResultsButton.vue
```diff
@@ -35,26 +47,57 @@ export default {
...
+          res.subject_data.forEach((subject) => {
             csvRows.push([
-              res.dataset_name,
-              res.num_matching_subjects,
-              path,
+              res.dataset_file_path,
+              subject.sub_id,
+              subject.age,
+              subject.sex,
+              subject.diagnosis?.join('  '),
```
surchs: This confuses me a little bit. Are you saying that if we get an array of diagnoses, that they should be concatenated with two spaces?

That seems tricky to use for me. Ideas:
- make the whole file a .tsv and then use comma as delimiter here
- force this into a string of an array, i.e. `"['hello', 'world']"`. Not great, but probably a bit easier to parse than just spaces

components/DownloadResultsButton.vue
```diff
@@ -35,26 +47,57 @@ export default {
...
+      const datasets = this.results.filter((res) => this.downloads.includes(res.dataset_name));
 
-      this.results.filter((res) => this.downloads.includes(res.dataset_name))
-        .forEach((res) => {
-          res.subject_data.forEach((path) => {
+      if (this.toggleCSV) {
+        const headers = ['DatasetFilePath', 'SubjectID', 'Age', 'Sex', 'Diagnosis', 'SessionID', 'SessionPath', 'NumSessions', 'Modality'];
+        csvRows.push(headers);
+
+        datasets.forEach((res) => {
```
surchs, line 72: :cherries: but since (if I understand right) you actually DO want the result of the iteration stored to a variable, why  not just use `.map` here and assign to `csvRows`?

components/QueryForm.vue
```diff
@@ -151,6 +153,7 @@ export default {
       }
     },
     async submitQuery() {
+      this.isFetching = true;
```
surchs, line 159: :tada:

surchs (approved):
Thanks for the PR @rmanaem. All looks very good, I only have minor :cherries: comments.
One is about naming `toggleCSV` something else in the code to help with understanding the functionality easier (I think the user facing side of the functionality is clear).

The other thing that I didn't comment: We really should implement a way to have a preview-deploy of the APP as it'll look and behave once the PR is in. Now that we have all of these great tests, I rarely checkout the code because I trust that it works. But you can't really write tests for the question of "how does it look like" - and that is also important.
Just brain dumping this, already mentioned it at the end of https://github.com/neurobagel/project/issues/107. I believe vercel is our best shot for this.

Anyhow. Good work, good to go :cook:

components/DownloadResultsButton.vue
surchs: Oh, hold on, now that I look at it again: I think this would be more appropriately done as a computed property, no?
As in: you don't want to call the function repeatedly here: https://v2.vuejs.org/v2/guide/computed.html In fact I'm not completely sure how this ends up being reactive as a method call.

components/DownloadResultsButton.vue
rmanaem: How about `toggleResultsTSV`?

components/DownloadResultsButton.vue
rmanaem: Yea, the two spaces are basically used as separators here.
I didn't wanna switch to t.sv since we have it as .csv everywhere but I'd be happy to switch.

components/DownloadResultsButton.vue
surchs: @jarmoza any thoughts here? To me a "toggle" is an on/off switch, but that's not important. important is only that whoever reads this next spends as little time as possible understanding what this variable does.

jarmoza: 'toggle' in code works and the verb is correct but I think the confusion here will be in the checkbox text itself `Toggle Results TSV`. It would be helpful to know what we're toggling between. A possible change to that text might remedy the issue.

Some questions that I can think of on the spot:

1. It's an either/or participant-level or dataset-level results?
2. Does the latter include the former?
3. Or would it even be helpful to just have two separate buttons?

rmanaem: The button's text gets updated when the checkbox is checked/unchecked showing which results tsv is ready for download.

PR #100, opened 2023-06-06
[BUG] Fixed team feedback bugs

### LINKED ISSUE #95: Query for 0 sessions throws an error
When a query with the value 0 for minimum number of session is submitted, the query tool throws a generic error. The error is generated by API see [here](https://github.com/neurobagel/api/blob/f891ff0860589e9fa5590d092d220934ae395597/app/api/models.py#LL18C54-L18C54).

### LINKED ISSUE #96: The dataset `MRI Lab Graz: Reading-related functional activity in children with isolated\nspelling deficits and dyslexia` break the format of the `dataset-results.tsv` file
The dataset name contains the `\n` escape character, it is parsed by JavaScript in code and interpreted as a new line when constructing the tsv file.

rmanaem:
Fixes neurobagel/old-query-tool#95 
Fixes neurobagel/old-query-tool#96 
See also neurobagel/query-tool#70 

- Implemented a condition to give an error message via toast when the value of the minimum number of sessions is set to 0 in `validateQueryForm` 
- Implemented a test case for the above in the `QueryForm` component test
- Implemented a quick fix for neurobagel/old-query-tool#96, to replace `\n` escape character in `dataset_name` with a space
- Implemented `DatasetResultsTSV` e2e test for the above

## Checklist

- [ ] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/DownloadResultsButton.vue
+++ b/components/DownloadResultsButton.vue
@@ -92,9 +92,9 @@ export default {
           tsvRows.push([
             res.dataset_portal_uri,
             res.dataset_file_path,
-            res.dataset_name,
+            res.dataset_name.replace('\n', ' '),
             res.num_matching_subjects,
-            res.image_modals.join(', '),
+            res.image_modals?.join(', '),
           ].join('\t'));
         });
       }
@@ -104,7 +104,7 @@ export default {
 
     downloadResults() {
       const element = document.createElement('a');
-      element.setAttribute('href', `data:text/tsv;charset=utf-8,
+      element.setAttribute('href', `data:text/tab-separated-values;charset=utf-8,
       ${encodeURIComponent(this.generateTSVString())}`);
       if (this.toggleResultsTSV) {
         element.setAttribute('download', 'participant-results.tsv');
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -150,7 +150,9 @@ export default {
     },
     validateQueryForm() {
       if (this.minAge && this.maxAge && Number(this.minAge) > Number(this.maxAge)) {
-        this.displayToast('Maximum age must be greater than or equal to minimum age');
+        this.displayToast('The value of maximum age field must be greater than or equal to the value of minimum age field');
+      } else if (this.min_num_sessions === '0') {
+        this.displayToast('The value for minimum number of sessions field must be greater than 0');
       } else {
         this.submitQuery();
       }
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -53,6 +53,7 @@
           </template>
         </categorical-field>
         <continuous-field
+          min="1"
           name="Minimum number of sessions"
           data-cy="min-num-sessions-field"
           step="1"
@@ -151,8 +152,6 @@ export default {
     validateQueryForm() {
       if (this.minAge && this.maxAge && Number(this.minAge) > Number(this.maxAge)) {
         this.displayToast('The value of maximum age field must be greater than or equal to the value of minimum age field');
-      } else if (this.min_num_sessions === '0') {
-        this.displayToast('The value for minimum number of sessions field must be greater than 0');
       } else {
         this.submitQuery();
       }
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -66,11 +66,5 @@ describe('Query form', () => {
     cy.get('[data-cy="submit-query"]').click();
     // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
     cy.contains('#b-toaster-top-right', 'The value of maximum age field must be greater than or equal to the value of minimum age field');
-
-    cy.get('[data-cy="Min Age-continuous-field-input"]').clear();
-    cy.get('[data-cy="Max Age-continuous-field-input"]').clear();
-    cy.get('[data-cy="Minimum number of sessions-continuous-field-input"]').type(0);
-    cy.get('[data-cy="submit-query"]').click();
-    cy.contains('#b-toaster-top-right', 'The value for minimum number of sessions field must be greater than 0');
   });
 });
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -65,6 +65,12 @@ describe('Query form', () => {
     cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
     cy.get('[data-cy="submit-query"]').click();
     // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
-    cy.contains('#b-toaster-top-right', 'Maximum age must be greater than or equal to minimum age');
+    cy.contains('#b-toaster-top-right', 'The value of maximum age field must be greater than or equal to the value of minimum age field');
+
+    cy.get('[data-cy="Min Age-continuous-field-input"]').clear();
+    cy.get('[data-cy="Max Age-continuous-field-input"]').clear();
+    cy.get('[data-cy="Minimum number of sessions-continuous-field-input"]').type(0);
+    cy.get('[data-cy="submit-query"]').click();
+    cy.contains('#b-toaster-top-right', 'The value for minimum number of sessions field must be greater than 0');
   });
 });
--- a/cypress/e2e/DatasetResultsTSV.cy.js
+++ b/cypress/e2e/DatasetResultsTSV.cy.js
@@ -1,0 +1,21 @@
+const response = [
+  {
+    dataset_portal_uri: 'https://someportal.org/datasets/ds0001',
+    dataset_file_path: 'https://github.com/somethingDatasets/ds0001.git',
+    dataset_name: 'some\nname',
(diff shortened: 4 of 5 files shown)


components/DownloadResultsButton.vue
```diff
@@ -104,7 +104,7 @@ export default {
 
     downloadResults() {
       const element = document.createElement('a');
-      element.setAttribute('href', `data:text/tsv;charset=utf-8,
+      element.setAttribute('href', `data:text/tab-separated-values;charset=utf-8,
```
surchs, line 107: interesting. was that incorrect before?

components/DownloadResultsButton.vue
```diff
@@ -92,9 +92,9 @@ export default {
           tsvRows.push([
             res.dataset_portal_uri,
             res.dataset_file_path,
-            res.dataset_name,
+            res.dataset_name.replace('\n', ' '),
```
surchs, line 95: like we said today: I think this is an appropriate fix for the specific problem. I also think that we should find a more generic way of addressing the problem without changing data that we pass to the user. E.g. by encapsulating everything in a string. Maybe we can add a TODO here with a link to the issue / PR to revisit this problem in the near future?

components/QueryForm.vue
```diff
@@ -150,7 +150,9 @@ export default {
     },
     validateQueryForm() {
       if (this.minAge && this.maxAge && Number(this.minAge) > Number(this.maxAge)) {
-        this.displayToast('Maximum age must be greater than or equal to minimum age');
+        this.displayToast('The value of maximum age field must be greater than or equal to the value of minimum age field');
+      } else if (this.min_num_sessions === '0') {
+        this.displayToast('The value for minimum number of sessions field must be greater than 0');
```
surchs: just to check: can we not set this as a constraint on the field itself (i.e. like we disallow negative values)?

surchs: hhm, I see `min` as an attribute to `<b-form-input>` in the `continuousField.vue` component. That seems to me as if we should be able to explicitly set the constraint there by having a local default that can be overridden by the calling parent, no?

cypress/component/QueryForm.cy.js
```diff
@@ -65,6 +65,12 @@ describe('Query form', () => {
     cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
     cy.get('[data-cy="submit-query"]').click();
     // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
-    cy.contains('#b-toaster-top-right', 'Maximum age must be greater than or equal to minimum age');
+    cy.contains('#b-toaster-top-right', 'The value of maximum age field must be greater than or equal to the value of minimum age field');
+
+    cy.get('[data-cy="Min Age-continuous-field-input"]').clear();
+    cy.get('[data-cy="Max Age-continuous-field-input"]').clear();
+    cy.get('[data-cy="Minimum number of sessions-continuous-field-input"]').type(0);
```
surchs: Those `data-cy` names don't seem right to me. Are you sure there are elements with these names? Or maybe I am missing something

cypress/component/QueryForm.cy.js
```diff
@@ -65,6 +65,12 @@ describe('Query form', () => {
     cy.get('[data-cy="Max Age-continuous-field-input"]').type(3);
     cy.get('[data-cy="submit-query"]').click();
     // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
-    cy.contains('#b-toaster-top-right', 'Maximum age must be greater than or equal to minimum age');
+    cy.contains('#b-toaster-top-right', 'The value of maximum age field must be greater than or equal to the value of minimum age field');
+
+    cy.get('[data-cy="Min Age-continuous-field-input"]').clear();
```
surchs: I like these tests, but I think we should start splitting them up. There is a lot going on here already with minimum and maximum age. Better to handle each field in its own test (i.e. 1: `"out of bounds age raises a toast UI element"` and 2: `"session number below 0 raises display toast"`). These tests should be short and with a single purpose

cypress/e2e/DatasetResultsTSV.cy.js
```diff
@@ -0,0 +1,21 @@
...
+    dataset_file_path: 'https://github.com/somethingDatasets/ds0001.git',
+    dataset_name: 'some\nname',
+    num_matching_subjects: 3,
+    subject_data: [],
+    image_modals: ['http://purl.org/nidash/nidm#FlowWeighted', 'http://purl.org/nidash/nidm#T1Weighted'],
+  },
+];
+
+describe('Dataset results TSV', () => {
+  it('Intercepts the request sent to the API and asserts over the request url', () => {
```
surchs: I think the test name should describe the behavioural process rather than the implementation. So "Removes a newline character from the downloaded results file if one is present in the data" or something like that.

cypress/e2e/DatasetResultsTSV.cy.js
```diff
@@ -0,0 +1,21 @@
...
+    cy.intercept('query/?*', response).as('call');
+    cy.visit('/');
+    cy.get('[data-cy="submit-query"]').click();
+    cy.get('[data-cy="select-all"]').check();
+    cy.get('[data-cy="download-results-button"]').click();
+    cy.readFile('cypress/downloads/dataset-results.tsv').should('not.contain', 'some\nname');
```
surchs: Very neat! However: the assertion would also be true if your code just nuked the `"some name"` string completely out of existence - which is not what we want. So I think it'd make more sense to assert that `"some name"` **does** exist as well (or only, if you know that it only occurs this once)

surchs (changes requested):
Thanks a lot for the PR @rmanaem! All looks very good, thanks for figuring out how to test these bugs! I requested changes mostly because I think we should disallow bad session-numbers (i.e. `0`) to be entered in the query-form in the first place - and because I think the test assertions should be looked over again.

components/DownloadResultsButton.vue
rmanaem: It wasn't the standard MIME type

components/QueryForm.vue
rmanaem: Are you saying add a `min` prop to the `continuousField.vue` component that overwrites the `min` attribute of the `<b-form-input`? I like that

components/DownloadResultsButton.vue
rmanaem: Could you create an issue with a short description of what you had in mind? I prefer not to have TODOs in code as they tend to go unnoticed.

cypress/component/QueryForm.cy.js
rmanaem: Yea, it's the `name` of the field which is also used as the label separated by `-`, concatenated with the type of field `continuous` or `categorical` separated by `-`, and concatenated by the type of html element

components/QueryForm.vue
surchs: yeah, I think so. If you make it a prop, then I'd make the default whatever it currently is (presumably `0`) to not accidentally break things

cypress/e2e/DatasetResultsTSV.cy.js
rmanaem: Are you referring to the `describe` or the `it`?

surchs: `it`

cypress/component/QueryForm.cy.js
rmanaem: Removing the part for the `minimum number of sessions` field now that we're using a prop.

surchs:
@rmanaem I don't think you commited any changes. Did you mean to push them still or do you want me to review just based on your comments?

rmanaem:
This is weird, see [branch history](https://github.com/neurobagel/query-tool/commits/bug-95-96)

rmanaem:
There we go

cypress/component/QueryForm.cy.js
```diff
@@ -66,11 +66,5 @@ describe('Query form', () => {
     cy.get('[data-cy="submit-query"]').click();
     // See https://stackoverflow.com/questions/71295432/unable-to-see-toast-in-cypress/71301155#71301155
     cy.contains('#b-toaster-top-right', 'The value of maximum age field must be greater than or equal to the value of minimum age field');
-
-    cy.get('[data-cy="Min Age-continuous-field-input"]').clear();
```
surchs: ok, cool. But now we don't have any test for the feature of "cannot enter < 1 into session field". That's not ideal either. Mind adding a test for that?

surchs (approved):
Looks good @rmanaem. Noticed that we now have no test for number-session  field. I think that'd be good to add. Otherwise good to go :cook:

cypress/component/QueryForm.cy.js
rmanaem: I don't think it's possible to test that. If you recall we talked about this when implementing `continuousField` component which at that time set the `min` attribute to 0 and there wasn't a way to test for that in cypress.

surchs: aha. good point. In that case: :cook: !

PR #105, opened 2023-06-08
[ENH]: add breakpoints

joshunrau:
- Add breakpoints to `QueryForm` and `ResultsContainer` so they will not break on mobile
- Change column sizes for large screens so the UI will not break on smaller "xl" screens. For example, on my 13" Macbook, the query panel is very small and the "Healthy control" checkbox overflows.
- Center the X to clear the categorical field so it is aligned with the dropdown indicator
- 
## Checklist

- [ ] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-col cols="12" md="4" xl="3">
+  <b-col cols="12" md="4" xl="2">
     <b-row>
       <h2>Query fields</h2>
     </b-row>
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -1,6 +1,5 @@
 <template>
-  <b-col cols="12" md="4" xl="3"
-  >
+  <b-col cols="12" md="4" xl="3">
     <b-row>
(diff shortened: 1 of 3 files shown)


rmanaem (changes requested):
Hey @joshunrau, thanks a lot for your first submission to Neurobagel! Happy to have you here :tada:, we appreciate your input. I think your suggestion to add breakpoints makes sense here, thanks for the PR. However: the query form columns with your PR are now quite a bit wider than they were before when seen on a normal-sized desktop viewport. Since we're optimizing viewing experience for larger screens, could you please adjust the changes to make the query fields less wide?

Let me know if you have time to address this, we'll be happy to merge with these changes

joshunrau:
Okay I changed it back. I would suggest looking into implementing an xxl breakpoint size since on a borderline xxl screen these breakpoints don't look very good imo

rmanaem:
Yea, the overall layout needs some work.

rmanaem (approved):
🧑‍🍳

PR #135, opened 2023-07-05
[BUG] Fixed inconsistent behaviour of checkboxes

### LINKED ISSUE #124: [BUG] Old checkbox state in result cards produces blank output file

### Expected Behavior

Initiating a successive query should produce result cards that are all unchecked, and additionally, downloaded results should have the appropriate subject- or dataset-level tsv data.

### Current Behavior

1. After initiating a query (i.e. querying all the datasets without filters) and selecting a few of the dataset results by clicking on their checkboxes, successive queries afterwards should have those checked checkbox states be cleared to unchecked.
2. When downloading results from such successive queries, proper results should be in the downloaded tsv file.

### Error message

_No response_

### Environment

_No response_

### How to reproduce

1. Query all
2. Select a few datasets in the results by clicking on their checkboxes
3. Alter the query filter(s)
4. Query again
5. Note that checkboxes in the new returned results remain checked
6. Attempt to download subject- or dataset-level datasets from the results
7. Note that the output file is blank except for the header row

### Anything else?

_No response_

### LINKED ISSUE #134: Datasets do not get removed from `downloads` array when new query is run
Once a dataset from the results is selected it is added to the `downloads` array and the only way to remove it from the `downloads` is to unselect it (uncheck the checkbox).
So when a query is run and dataset X from the results is selected (therefore added to the `downloads` array) but not unselected (not removed from the `downloads`) before running another query which may or may not include dataset X in the returned results, X is still included in the `downloads` array.

rmanaem:
Closes #124 
Closes #134 

- Renamed `selectAll` prop to `isChecked` in the `ResultCard` component
- `updateDownloads` method now uses `event` instead of `refs` in the `ResultCard` component
- Changed the `v-for` `key` to `dataset_uuid` in the `ResultsContainer` component
- Modified the value of the `is-checked` prop passed to `ResultCard` in `ResultsContainer` to check if the `datasetName` is already in the `downloads` array or is `selectAll` set to `true`
- Implemented an integration test to ensure the `downloads` array is wiped every time a query is run
  - documented here in [wiki](https://github.com/neurobagel/documentation/wiki/Cypress-Test-Snippets#provide-different-responses-based-on-a-condition-when-intercepting-requests)
- Implemented a watcher in `ResultsContainer` to wipe the `downloads` array every time `results` is changed i.e. a new query is submitted
- Updated `ResultCard` component test

## Checklist

- [x] Tests pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -112,6 +112,12 @@ export default {
       return !Object.is(this.results, null) && this.results.length !== 0;
     },
   },
+  watch: {
+    results() {
+      this.downloads = [];
+    },
+
+  },
   methods: {
     summarizeStats() {
       let datasets = 0;
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -71,11 +71,11 @@
       <b-list-group id="results-list-group">
         <result-card
           v-for="res in results"
-          :key="res.id"
+          :key="res.dataset_uuid"
           :dataset-name="res.dataset_name"
           :num-matching-subjects="res.num_matching_subjects"
           :image-modals="res.image_modals"
-          :select-all="selectAll"
+          :is-checked="selectAll"
           :data-cy="res.dataset_name"
           @update-downloads="updateDownloads"
         />
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -114,6 +114,7 @@ export default {
   },
   watch: {
     results() {
+      this.selectAll = false;
       this.downloads = [];
     },
 
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -75,7 +75,7 @@
           :dataset-name="res.dataset_name"
           :num-matching-subjects="res.num_matching_subjects"
           :image-modals="res.image_modals"
-          :is-checked="selectAll"
+          :is-checked="selectAll || downloads.includes(res.dataset_name)"
           :data-cy="res.dataset_name"
           @update-downloads="updateDownloads"
         />
--- a/cypress/e2e/Checkbox.cy.js
+++ b/cypress/e2e/Checkbox.cy.js
@@ -30,6 +30,8 @@ const response2 = [
 
 describe('Checkbox', () => {
   it('Unchecks checked checkboxes after a second query is run.', () => {
+    // Since the defualt viewport results in partial occlusion of the checkbox
+    // Set the viewport to 2000x1000 to ensure the checkbox is not covered and is cliclable
     cy.viewport(2000, 1000);
 
     let isFirstClick = true;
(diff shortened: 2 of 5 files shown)


components/ResultsContainer.vue
```diff
@@ -112,6 +112,12 @@ export default {
       return !Object.is(this.results, null) && this.results.length !== 0;
     },
   },
+  watch: {
+    results() {
```
surchs, line 116: I have mixed feelings here. This is short and works. But there is also a more direct way to figure out if a query was run for us. I.e. "did user click 'query' button". This information is not accessible here, because it's handled by the parent (? or the query form component). But maybe that means the component handling the queries should also be responsible for the `downloads` array. 

I can think of two issues with this solution:
1. If I run the exact same query twice, datasets will (probably ?) not get wiped because `results` are the same (?) - mild inconsistency
2. If we ever start doing what you suggested and manipulate `results` after receiving it from the API (e.g. to manage state), we might be sitting there scratching our heads why all the datasets got wiped.

Neither of them is important, so I'd say let's keep it here. But this also indicates that we may have to think about state management a bit more for this app.

cypress/e2e/Checkbox.cy.js
```diff
@@ -0,0 +1,52 @@
...
+    dataset_portal_uri: 'https://openneuro.org/datasets/ds001419',
+    num_matching_subjects: 3,
+    subject_data: [],
+    image_modals: ['http://purl.org/nidash/nidm#FlowWeighted', 'http://purl.org/nidash/nidm#T1Weighted'],
+  },
+];
+
+describe('Checkbox', () => {
+  it('Unchecks checked checkboxes after a second query is run.', () => {
+    cy.viewport(2000, 1000);
```
surchs, line 35: Can you add a comment as to why this is being set here. I remember it's about sizing of viewport resulting in partial occlusion of elements. 

Also: can you open an issue (and link here) about the partial occlusion problem

surchs (approved):
Very nice PR, thanks @rmanaem. All changes look good to me, but I have some comments on the watcher. Take a look, but this shouldn't hold the PR :cook: 

Sidenote: The preview is nice, but it isn't super useful for this app, because the app gets deployed without a working backend to provide any query results. I wonder if we should just have a `test.neurobagel.org` graph that is small and always on as a solution for all of our testing purposes (at least until we have our `docker compose up` dev envs)

cypress/e2e/Checkbox.cy.js
rmanaem: See https://github.com/neurobagel/query-tool/issues/136

components/ResultsContainer.vue
rmanaem: I understand your concern.

> If I run the exact same query twice, datasets will (probably ?) not get wiped because results are the same (?) - mild inconsistency

This doesn't happen. It properly wipes the `downlaods` array even if the `results` are the same.

I noticed another issue, when `selectAll` is checked and another query is run, all datasets are checked but `downloads` is wiped. I resolved this by setting `selectAll` to false along with wiping the `downloads`.

Issue #117, opened 2023-06-12
Sketch the design of the navbar


rmanaem:
Create a sketch for the proposed design of the navbar of the query tool based on shortcuts mentioned in #103 
See [Bootstrap-vue](https://bootstrap-vue.org/) as inspiration

rmanaem:
@neurobagel/dev What do you think?

![Image](https://github.com/neurobagel/query-tool/assets/77515879/0799abd8-bffc-454e-aad5-5c8fa85ae2d6)

surchs:
I think that looks absolutely awesome @rmanaem! I'm for it

surchs:
@rmanaem I think we can have this issue in review now, wdyt?

alyssadai:
Ou, I love the idea of the dropdown with different options for feedback!

I'm still slightly attached to the idea of having some kind of indicator of what "stage" of the Neurobagel workflow folks are on (annotate, query, etc.). Since we won't include the tool name in the navbar, what do we think of adding a descriptive title to the page similar to [OpenNeuro's search interface](https://openneuro.org/search?query={%22keywords%22:[]}), maybe with something like "Search across Neurobagel-annotated datasets" or something similar? :)

Could of course be a separate issue.

rmanaem:
> what do we think of adding a descriptive title

Like in the tab title?
I like the idea of having the tool mentioned, so how about a `tools/services` shortcut that has a dropdown similar to the feedback shortcut.

alyssadai:
Not a lab title, but a title on each tool page that summarizes the tool's purpose.

So for example, if a person heads directly to `query.neurobagel.org`, they don't just see the query fields, but also a title on the page that says e.g. "Search across Neurobagel-annotated datasets", so they know what it is they're actually doing when they fill out the query form. Similarly, for the annotation tool, the title could say "Generate harmonized phenotypic annotations of dataset" or something.

surchs:
@rmanaem could you update your proposal based on the feedback (or explain why you disagree) and then close this issue please?

rmanaem:
@alyssadai Is this what you had in mind (font size to be adjusted)?

![Image](https://github.com/neurobagel/query-tool/assets/77515879/c989eb58-2d49-4617-ad52-cebf2f3d1a37)

alyssadai:
[...]

PR #142, opened 2023-07-11
[ENH] Navbar

### LINKED ISSUE #103: [ENH] Add shortcuts
Implement the following shortcuts in the tool for users to have easy access to

- [x] #117
- [x] Feedback shortcut to GitHub issues of the repo
- [x] Website shortcut
- [x] Repo shortcut
- [x] Documentation shortcut

rmanaem:
Closes #103 
See also #117

-  Added default layout
- Implemented Navbar component
- Implemented component test for Navbar
- Removed the `Query Tool` title from the `index.vue`
- Modified the style of buttons
- Instructed eslint to ignore `vue/multi-word-component-names` rule for the `default.vue` and `Navbar.vue`

## Checklist

- [x] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:


surchs:
checking the staging build (:tada: btw), the stuff on the right seems to be centered on both the title and subtitle. I think centered on just the title might look better:
image

components/Navbar.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div>
+    <b-navbar
+      toggleable="lg"
+      type="light"
+      variant="light"
+    >
+      <b-navbar-brand
```
jarmoza, line 8: I think just a small space beneath the navbar brand, like 0.5em to separate navbar from app body.

components/Navbar.vue
```diff
@@ -0,0 +1,61 @@
...
+        class="brand-styling"
+        href="https://www.neurobagel.org/"
+        data-cy="brand"
+      >
+        <b-row>
+          Neurobagel Query
+        </b-row>
+        <b-row id="nav-text">
+          define and find cohorts at the sample level
```
jarmoza: Might want to consider placing text in the component data. But I would say definitely capitalize 'd' in 'define' and make sure the subtitle looks aligned with the title.

components/QueryForm.vue
```diff
@@ -76,7 +76,7 @@
           @update-categorical-field="updateField"
         />
         <b-button
-          variant="primary"
+          class="nb-button"
```
jarmoza, line 79: Just a note as we consider palette alignment in the future, the active button color in the annotation tool is currently green.

pages/index.vue
```diff
@@ -105,87 +99,125 @@ export default {
 };
 </script>
 <style>
-/* Card */
 
-.btn-group {
+/* Navbar */
+.brand-styling {
```
jarmoza: It also might be time to consider moving global styles out of this file and into a separate css file.
The annotation tool currently has an `assets/css` folder with a `main.css` inside.

This can be imported by going to `nuxt.config.js` and adding this:

```
	// Global CSS: https://go.nuxtjs.dev/config-css
	css: [

		"@/assets/css/main.css"
	],
```

components/Navbar.vue
```diff
@@ -0,0 +1,61 @@
...
+      >
+        <b-navbar-nav
+          id="right-nav"
+          class="ml-auto"
+        >
+          <b-nav-item
+            href="https://github.com/neurobagel/query-tool/tree/v0.1.0/"
+            data-cy="version"
+          >
+            v0.1.0
```
jarmoza, line 35: For all of these new links, including the link to neurobagel.org, we might want to consider adding a `target="_blank"` in order to open up a new tab as this may be less disruptive for users.

What do you think, @rmanaem ?

cypress/component/Navbar.cy.js
```diff
@@ -0,0 +1,29 @@
+import Navbar from '../../components/Navbar.vue';
+
+describe('Navbar', () => {
+  it('Displays the navbar with the correct links', () => {
+    cy.mount(Navbar);
+    cy.get('[data-cy="brand"]').should('be.visible');
+    cy.get('[data-cy="brand"]').contains('Neurobagel Query');
```
jarmoza, line 7: Add a test for the subtitle now?

jarmoza (changes requested):
A few light change requests and suggestions.

components/Navbar.vue
rmanaem: > make sure the subtitle looks aligned with the title.

Could you elaborate?

rmanaem:
>  the stuff on the right seems to be centered on both the title and subtitle. I think centered on just the title might look better


@surchs  Yea, that's because the subtitle and the title are both in the same `nav-brand` component.
It's doable but might take some time and since we are a bit time-constrained, I think it's better to leave it to another time.
Feel free to open an issue so we can address it later.

components/Navbar.vue
jarmoza: The more text we place directly into the app, the more it becomes useful to keep the strings in a more central place. In the annotation tool, my current practice is to include a variable `uiText` in the component's `data` object.

Ex.
```
                uiText: {

                    instructions: "Annotate each unique value",
                    missingValueButton: "Mark as missing",
                    saveButton: "Save Annotation"
                }
```

As to the latter: `Define and find cohorts at the sample level`

rmanaem: I was referring to this part of your comment:

> make sure the subtitle looks aligned with the title.

surchs:
> > the stuff on the right seems to be centered on both the title and subtitle. I think centered on just the title might look better
> 
> @surchs Yea, that's because the subtitle and the title are both in the same `nav-brand` component. It's doable but might take some time and since we are a bit time-constrained, I think it's better to leave it to another time. Feel free to open an issue so we can address it later.

Good call. Don't waste time on this! I'll make a note

components/Navbar.vue
jarmoza: Oh! When I was looking at the left alignment of the title and subtitle, the subtitle appeared a little bit to the left of the title (in alignment). Could just be the typeface though and not the box sizing.

jarmoza (approved):
🧑‍🍳

PR #159, opened 2023-07-17
[BUG] Modified components to use `dataset_uuid` as identifier

### LINKED ISSUE #153: [BUG] Duplicate datasets in OpenNeuro are selected simultaneously when one is checked
To replicate:

Query for a "Depressive disorder" diagnosis. Among the results will be two datasets with the same name ("Two sessions of resting state with closed eyes..."), and checking the box for one selects both.

These two datasets are 'true' duplicates in OpenNeuro, i.e. they have separate dataset IDs (`dsXXXXXX`) and repos. So checking off one dataset should not check the other dataset as well.

rmanaem:
Closes #153 

- Implemented `datasetId` prop as an identifier for datasets
- Modified `data-cy` attribute of elements to use `datasetId`
- Updated `updateDownloads` method
- Updated `updateDownloadsSelectAll` method
- Modified `generateTSVString` method to use `dataset_uuid` for filtering the `results` array
- Updated tests

## Checklist

- [ ] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultCard.vue
+++ b/components/ResultCard.vue
@@ -15,7 +15,7 @@
                   type="checkbox"
                   value=""
                   :checked="isChecked"
-                  :data-cy="`card-${datasetName}-checkbox`"
+                  :data-cy="`card-${datasetId}-checkbox`"
                   @change="updateDownloads"
                 >
               </b-col>
@@ -26,7 +26,7 @@
                 <b-row>
                   <h5
                     class="card-title"
-                    :data-cy="`card-${datasetName}-dataset`"
+                    :data-cy="`card-${datasetId}-dataset`"
                   >
                     {{ datasetName }}
                   </h5>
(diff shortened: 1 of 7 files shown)


components/ResultCard.vue
```diff
@@ -15,7 +15,7 @@
                   type="checkbox"
                   value=""
                   :checked="isChecked"
-                  :data-cy="`card-${datasetName}-checkbox`"
+                  :data-cy="`card-${datasetId}-checkbox`"
```
surchs: datasetUUID? to keep it consistent

components/ResultCard.vue
```diff
@@ -74,6 +74,10 @@
 <script>
 export default {
   props: {
+    datasetId: {
```
surchs: `UUID`.

surchs (approved):
Look good @rmanaem. One request: let's change the propName to `datasetUUID` to save ourselves some confusion down the road (i.e. that this is about the UUID and not e.g. the OpenNeuro id)

surchs:
And with that good to go :cook:

components/ResultCard.vue
rmanaem: it would have to be `datasetUuid` Vue doesn't seem to like all caps `UUID`

PR #161, opened 2023-07-18
[BUG] Removed the unwanted whitespace from results tsv files

### LINKED ISSUE #151: [BUG] Dataset and Participant .tsv downloads have `\n` when downloaded from Chrome, but not Firefox
When I
- run a query
- select datasets
- click the participant or dataset download button, 

then I expect:
### Expected Behavior

To get a `.tsv` that starts with a header row, e.g.: 

```tsv
DatasetID	SubjectID	Age	Sex	...
http://neurobagel.org/vocab/0143b0cc...	sub-s102	19.0	http://purl.bioontology.org/ontology/SNOMEDCT/248153007
```

### Current Behavior
But instead I get:
```tsv
\n
      DatasetID	SubjectID	Age	Sex	...
http://neurobagel.org/vocab/0143b0cc...	sub-s102	19.0	http://purl.bioontology.org/ontology/SNOMEDCT/248153007
```

### How to reproduce

- use the query app with Chrome (e.g. `114.0.5735.198`)

### Anything else?

- if you use query app with Firefox (e.g. `115.0`), there is no newline character in the .tsv :shrug: 

Whatever the cause of the bug is, we need to:

- [ ] write a test that fails when there is a newline at the top of the `.tsv`
- [ ] enable cross-browser testing (at least for cypress) or at least make sure tests are run on Chrome (as sad as it is, most people are on Chrome)

rmanaem:
Closes #151 

- Refactored `downloadResults` method
- Implemented a test for the bug
- Renamed `DatasetResultsTSV.cy.js` to `ResultsTSV.cy.js`

## Checklist

- [x] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/e2e/ResultsTSV.cy.js
+++ b/cypress/e2e/ResultsTSV.cy.js
@@ -9,7 +9,7 @@ const response = [
   },
 ];
 
-describe('Dataset results TSV', () => {
+describe('Results TSV', () => {
   it('Removes a newline character from a dataset name in the downloaded dataset-level results file', () => {
     cy.viewport(2000, 1000);
     cy.intercept('query/?*', response).as('call');
--- a/cypress/e2e/ResultsTSV.cy.js
+++ b/cypress/e2e/ResultsTSV.cy.js
@@ -26,8 +26,12 @@ describe('Results TSV', () => {
     cy.get('[data-cy="submit-query"]').click();
     cy.get('[data-cy="select-all"]').check();
     cy.get('[data-cy="download-dataset-level-results-button"]').click();
-    cy.readFile('cypress/downloads/dataset-results.tsv').should('not.contain', '\n      DatasetID');
+    cy.readFile('cypress/downloads/dataset-results.tsv').then((fileContent) => {
+      expect(fileContent).to.match(/^DatasetID/);
+    });
     cy.get('[data-cy="download-participant-level-results-button"]').click();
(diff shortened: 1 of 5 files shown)


surchs (approved):
:cook: :tada:

cypress/e2e/ResultsTSV.cy.js
```diff
@@ -19,4 +19,15 @@ describe('Dataset results TSV', () => {
...
     cy.readFile('cypress/downloads/dataset-results.tsv').should('contain', 'some name');
   });
+  it('Removes the unwanted whitespace from the downloaded results files', () => {
+    cy.viewport(2000, 1000);
+    cy.intercept('query/?*', response).as('call');
+    cy.visit('/');
+    cy.get('[data-cy="submit-query"]').click();
+    cy.get('[data-cy="select-all"]').check();
+    cy.get('[data-cy="download-dataset-level-results-button"]').click();
+    cy.readFile('cypress/downloads/dataset-results.tsv').should('not.contain', '\n      DatasetID');
```
surchs: I like this! But this is a bit brittle I think. Instead of saying "should not have newline followed by 6 whitespaces followed by ..." we should say: "should start with DatasetID" or, even better, "should start with a non-whitespace character". Otherwise this test will pass if you add two newlines -> :( or if you add seven spaces -> :( ...

surchs (changes requested):
thanks @rmanaem, looks great. Just one request to change the test logic. Then good

surchs (approved):
:tada: :cook:

PR #175, opened 2023-08-07
[MAINT] Refactored `DownloadResultsButton` into a generic component

### LINKED ISSUE #146: Refactor the `DownloadResultsButtton` component
Currently the `DownloadResultsButton` component houses two `b-button` elements one for participant-level results and one for dataset-level results.
Refactor the component to a generic one button component that receives its data from the parent through props.

rmanaem:
Closes #146 

- Refactored `DownloadResultsButton` into a generic component
  - Removed `b-col`, `b-row`, `div` containers
  - Removed duplicate code
  - Removed `displayDownloadResultsButton` computed property
  - Implemented `identifier` prop
  - Updated `downloadResults` method
- Updated `ResultsContainer` component
    - Implemented a second `download-results-button` to the template
    - Implemented container elements for `download-results-button` elements
- Updated styles
- Updated related tests
- Secondary refactors (after 1st review):
  - Moved `generateTSVString` and `downloadResults` methods from
`DownloadResultsButton` to `ResultsContainer`
  - Removed `results` and `downloads` props in `DownloadResultsButton`
  - Added `downloadIsEmpty` prop in `DownloadResultsButton`
  - Removed `displayToolTip` method and moved the logic in line
  - Implemented `download-results` event in `DownloadResultsButton`
  - Updated `DownloadResultsButton` component test
  - Added a test case for `download-results` event

## Checklist

- [x] Tests pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -99,7 +99,7 @@
   background-color: #2d0644;
 }
 
-.download-dataset-level-results-button {
+.download-buttons-div > *:not(:first-child) {
   margin-left: 0.25em;
 }
 
--- a/cypress/component/DownloadResultsButton.cy.js
+++ b/cypress/component/DownloadResultsButton.cy.js
@@ -1,6 +1,7 @@
 import DownloadResultsButton from '../../components/DownloadResultsButton.vue';
 
 const props = {
+  identifier: 'participant-level',
   results: [
     {
       dataset: 'http://neurobagel.org/vocab/cool-dataset',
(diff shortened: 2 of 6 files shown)


assets/css/main.css
```diff
@@ -99,7 +99,7 @@
   background-color: #2d0644;
 }
 
-.download-dataset-level-results-button {
+.download-buttons-div > *:not(:first-child) {
```
surchs, line 102: neat

surchs (approved):
Nice changes @rmanaem. This components are definitely much cleaner. I think this is good to merge as is :cook: 

But since you're already working on it, I think it would be even better if the responsibility for "make the download string" and "send the right download string" was also refactored into the `ResultsContainer.vue` component. 

Then the button component would be truly generic and just concerned with whether it was clicked (in response to which it should emit some helpful event that the `ResultsContainer` can respond to by doing the whole `generateTSVString` logic that currently lives insid the buttons. We might even be able to directly use one of the existing generic vue-bootstrap button components and remove the need for our own button.

The other benefit is that we don't have to push all the required information about downloads and results to the button via props - because the `ResultsContainer` already has all that info. 

Take a look and see if that's a quick refactor - otherwise let's discuss it in a new issue and then deal with it when it comes up again.

rmanaem:
@surchs I refactored the `DownloadResultsButton` and `ResultsContainer` components further. Let me know what you think.

cypress/component/DownloadResultsButton.cy.js
```diff
@@ -1,44 +1,42 @@
...
-      image_modals: ['http://purl.org/nidash/nidm#T1Weighted', 'http://purl.org/nidash/nidm#T2Weighted'],
-    },
-    {
-      dataset: 'http://neurobagel.org/vocab/not-so-cool-dataset',
-      dataset_name: 'not-so-cool-dataset',
-      num_matching_subjects: 2,
-      subject_file_paths: ['not-so-cool-path', 'some-not-so-cool-path'],
-      image_modals: ['http://purl.org/nidash/nidm#FlowWeighted', 'http://purl.org/nidash/nidm#T1Weighted'],
-    },
-
```
surchs, line 19: This is great! Just to make sure: do these props have to end up in `ResultsContainer.cy.js` now?

surchs (approved):
Looks great @rmanaem. Ready to go :cook: !

cypress/component/DownloadResultsButton.cy.js
rmanaem: Yea `ResultsContainer` component needs `results` as a prop.

PR #186, opened 2023-08-17
[ENH] Implemented example usage modal

### LINKED ISSUE #173: Code snippet on UI on how to download data
As a query user, having run my query, I want to be shown a piece of code I can copy/paste (single command ideally) that I can run so that my selected cohort results appear on my disk.

This is only about the UI element, we can figure out what the code is separately.
@rmanaem

rmanaem:
Closes #173

- Refactored `ResultsContainer` component to implement `Example usage` button
- Added `vue-js-modal` as a plugin
- Implemented `Example usage` button and modal
- Implemented test case for `Example usage` button  and modal
- Set the default viewport for cypress component and e2e tests in `cypress.config.js`
   - Updated related tests

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -93,7 +93,7 @@
             data-cy="example-usage-button"
             @click="$modal?.show('example-usage-modal')"
           >
-            Example usage
+            How to get the data
           </b-button>
           <modal
             name="example-usage-modal"
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -82,13 +82,38 @@
         />
       </b-list-group>
     </b-row>
-    <b-col
+    <b-row
       v-if="displayResults"
-      class="d-flex flex-row-reverse"
-      style="margin-top: 1em;"
+      style="margin-top: 1em"
     >
-      <b-row>
-        <div class="download-buttons-div d-flex">
+      <b-col cols="5">
+        <div class="d-flex justify-content-start">
+          <b-button
+            class="nb-button"
+            @click="$bvModal.show('example-usage-modal')"
+          >
+            Example usage
+          </b-button>
+          <b-modal
+            id="example-usage-modal"
+            title="Example usage"
+            hide-header-close
+            hide-footer
+            size="xl"
+          >
+            <p>Please follow these steps</p>
+            <ol>
+              <li>Select at least one dataset</li>
+              <li>Download the participant-level and dataset-level results</li>
+              <li>Change directory to the location of the downloaded files</li>
+              <li>Copy and run the following command:</li>
+            </ol>
+            <code class="code">{{ code }}</code>
+          </b-modal>
+        </div>
+      </b-col>
+      <b-col cols="7">
+        <div class="d-flex justify-content-end download-buttons-div">
           <download-results-button
             identifier="participant-level"
             :downloads-is-empty="downloads.length === 0"
@@ -102,8 +127,8 @@
             @download-results="downloadResults"
           />
         </div>
-      </b-row>
-    </b-col>
+      </b-col>
+    </b-row>
   </b-col>
 </template>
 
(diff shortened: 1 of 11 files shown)


components/ResultsContainer.vue
```diff
@@ -82,13 +82,43 @@
...
-      <b-row>
-        <div class="download-buttons-div d-flex">
+      <b-col cols="5">
+        <div class="d-flex justify-content-start">
+          <b-button
+            class="nb-button"
+            data-cy="example-usage-button"
+            @click="$modal?.show('example-usage-modal')"
+          >
+            Example usage
```
surchs: How about
```suggestion
            How to get the data
```
or something like that. Example usage sounds like you mean "... of the query tool"

cypress.config.js
```diff
@@ -2,13 +2,17 @@ const { defineConfig } = require("cypress");
 
 module.exports = defineConfig({
   component: {
+    viewportWidth: 2000,
```
surchs: I'm all for setting this in the config. But 2000 x 1000 is a pretty unusual size, and also pretty wide. How about 1280 x 1024 instead? Or something more typical

cypress/component/ResultsContainer.cy.js
```diff
@@ -97,4 +96,18 @@ describe('Results', () => {
     cy.get('[data-cy="download-participant-level-results-button"]').should('be.disabled');
     cy.get('[data-cy="download-dataset-level-results-button"]').should('be.disabled');
   });
+  it('Displays the example usage button and modal and checks the modal content', () => {
```
surchs: test name should have some reference to "should be visible only when results are displayed" - unless I misunderstand

cypress/component/ResultsContainer.cy.js
```diff
@@ -97,4 +96,18 @@ describe('Results', () => {
...
+    cy.mount(ResultsContainer, {
+      stubs,
+      propsData: props,
+    });
+    cy.get('[data-cy="example-usage-button"]').should('be.visible').click();
```
surchs, line 112: as above: would make sense to test that it isn't visible before the right conditions are met

cypress/component/ResultsContainer.cy.js
```diff
@@ -97,4 +96,18 @@ describe('Results', () => {
...
+    cy.get('[data-cy="example-usage-modal"]').should('be.visible');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'Please follow the steps below');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'Select at least one dataset');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'Download the participant-level and dataset-level results files');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'Change directory to the location of the downloaded files');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'Copy and run the following command:');
+    cy.get('[data-cy="example-usage-modal"]').should('contain', 'docker run -t -v $(pwd):/data neurobagel/dataget:latest /data/dataset-results.tsv /data/participant-results.tsv /data/output');
```
surchs: don't think this needs to be so specific. In fact that's a downside. The test should be concerned with the UI elements and behaviour - not with the exact content of the text in the component. We don't want the test to fail just because you corrected a typo in the instruction text

nuxt.config.js
```diff
@@ -31,6 +31,7 @@ export default {
   plugins: [
     '@plugins/bootstrap-vue',
     '@plugins/vue-select',
+    { src: '@plugins/vue-js-modal', ssr: false },
```
surchs, line 34: why is ssr false here? Mostly curiosity

surchs (approved):
looks good @rmanaem. I think the test needs another look over - I left some comments. Otherwise good to go :cook:

PR #199, opened 2023-08-25
[FIX] Correct usage modal with path names

### LINKED ISSUE #198: [BUG] "How to get data" tooltip has incorrect file names

### Expected Behavior

The docker command shown in the modal when clicking "How to get data" can be run as is

### Current Behavior

`docker run -t -v $(pwd):/data neurobagel/dataget:latest /data/dataset-results.tsv /data/participant-results.tsv /data/output`

is incorrect because the (default) names of the downloaded files are different (i.e. `participant-level-results.tsv` and `dataset-level-results.tsv`). See:

https://github.com/neurobagel/query-tool/blob/566524a828e637b3c333a2dab332f29d653f3771/components/ResultsContainer.vue#L241-L251

### Error message

_No response_

### Environment

_No response_

### How to reproduce

_No response_

### Anything else?

Thanks @ericearl for the report!

surchs:
- fixed path names on the modal
- added the names for the dataset and participant files to the test

Closes: #198

### DIFF
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -113,6 +113,7 @@
                 <li>Copy and run the following command:</li>
               </ol>
               <code class="code">{{ code }}</code>
+              <p>Note that the output directory will be created if it does not exist already.</p>
             </div>
           </modal>
         </div>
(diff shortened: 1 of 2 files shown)


components/ResultsContainer.vue
```diff
@@ -113,6 +113,7 @@
                 <li>Copy and run the following command:</li>
               </ol>
               <code class="code">{{ code }}</code>
+              <p>Note that the output directory will be created if it does not exist already.</p>
```
alyssadai:
```suggestion
              <p>The downloaded data will be saved in a directory called <code>output</code> in the current directory. Note that the output directory will be created if it does not exist already.</p>
```

alyssadai (approved):
One 🍒 for clarity, and good to go!

PR #204, opened 2023-09-01
[CI] Updated workflows to trigger on push to `main` branch

### LINKED ISSUE #203: CI isn't running on merge to main
We just merged https://github.com/neurobagel/query-tool/pull/199 with a failed linter check (not sure how tbh, we have branch protection on). Obviously linter check isn't breaking things - so no major harm done. But the main branch remained green after this. This made me realize that we aren't running **any** checks on push/merge to the main branch here - we only run the deployment. 

That's no good. If instead of a linter we had ignored a failed test then main should not be looking green anymore (and we also shouldn't deploy). If you break main, you need to know this (and then fix it right away). 

I think all tests **have to** run on main, but the linter **should** run because why not. 

We should then also confirm that if for some reason we push broken stuff to main and the tests fail, that we don't deploy!

rmanaem:
Closes #203 

- Updated component test, e2e test, and lint workflows to also trigger on push to `main` branch

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/component-test.yaml
+++ b/.github/workflows/component-test.yaml
@@ -1,6 +1,11 @@
 name: component tests
 
-on: [pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   component-test:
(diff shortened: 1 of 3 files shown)


surchs (approved):
Nice pr @rmanaem  thanks for fixing this so quickly 🧑‍🍳

PR #217, opened 2023-09-13
[MNT] Update dx and assessment tool dropdown options

### LINKED ISSUE #215: Update query tool to add Douglas early psychosis dataset DX and assessments
DX:
```bash
{
   "Label": "Acute schizophrenic episode",
   "TermURL": "snomed:268617001"
},
{
   "Label": "Family history of psychotic illness",
   "TermURL": "snomed:266969002"
},
```

Assessment tools:
```bash
{
    "TermURL": "cogatlas:trm_4f2411c91ae5e",
    "Label": "recall test"
}
{
     "TermURL": "cogatlas:trm_4c8a84f20dde2",
     "Label": "visual attention task"
}
{
     "TermURL": "cogatlas:trm_558702243da12",
     "Label": "Young Mania Rating Scale"
}
```

https://github.com/neurobagel/query-tool/blob/8cdc5f2a2e62d8b98f352807f253b9c22fe03cf8/pages/index.vue#L29-L45

### LINKED ISSUE #216: Add note to use https protocol when sending requests to `api.neurobagel.org`
If a user accidentally (🤡) sets `API_QUERY_URL` to `http://api.neurobagel.org/` instead of `https://api.neurobagel.org/` for a locally deployed query tool, they will get a CORS error even if the remote API is allowing all origins which is confusing to troubleshoot (or at least, the problem is not immediately obvious).

This is expected due to how NGINX is currently set up on the server: the API should only listen to https requests.

To avoid confusion, we should make it extra obvious in the README that if the user wants to use the remote API, they should use `https` and not `http`.

alyssadai:
Closes #215
Closes #216

- New terms added from Douglas early psychosis dataset annotation
- Added README note on using the correct protocol for `API_QUERY_URL`

## Checklist

- [x] Tests pass
- [x] Checks pass


surchs (approved):
:cook: thanks for the fix @alyssadai! 

Before you close #216 with this, take another look and check that this was in fact a CORS issue. If it was, I think we need to do a little more digging because that would be unexpected (at least by me).

PR #235, opened 2023-10-06
[CI] Fixed the job identifier for `build_docker` workflow

rmanaem:
## Checklist

- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/build_docker.yaml
+++ b/.github/workflows/build_docker.yaml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build docker:
+  build-docker:
    runs-on: ubuntu-latest
    steps:
     - name: Checkout


surchs (approved):
:cook: 

from https://github.com/neurobagel/query-tool/actions/runs/6470821514
I see that
```
[Invalid workflow file: .github/workflows/build_docker.yaml#L10](https://github.com/neurobagel/query-tool/actions/runs/6470821514/workflow)
The workflow is not valid. .github/workflows/build_docker.yaml (Line: 10, Col: 3): The identifier 'build docker' is invalid. IDs may only contain alphanumeric characters, '_', and '-'. IDs must start with a letter or '_' and and must be less than 100 characters.
```

I couldn't find a spec that says the same (and afaik YAML allows spaces in keys). Could you find a link to the spec or make a reference to no-spaces-in-job names in our internal docs please?

rmanaem:
@surchs I also couldn't find anything mentioning that. There was no mention of it in the GitHub docs for [wokflow syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions). I did make a short entry [here](https://github.com/neurobagel/documentation/wiki/Tools#workflow-file-syntax) in our wiki.

PR #269, opened 2023-11-09
[ENH] Updated the modal

### LINKED ISSUE #219: Enhancements for modal for downloading data
- [x] The modal currently is a bit too small which gives the appearance that text is being cut off (even though it's not). We should either resize the modal (not sure if you could do this dynamically depending on the amount of text?) or add a scrollbar.
- [x] Add a disclaimer that our command for automatically getting the data currently only applies to datasets available through datalad

rmanaem:
Closes #219 

- Set `height` and `width` of the modal manually
- Added a disclaimer for the command only working for datasets available through datalad

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -97,7 +97,8 @@
           </b-button>
           <modal
             name="example-usage-modal"
-            :height="auto"
+            :height="400"
+            :width="650"
             :adaptive="true"
             :resizable="true"
             :reset="true"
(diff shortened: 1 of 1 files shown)


components/ResultsContainer.vue
```diff
@@ -97,7 +97,8 @@
           </b-button>
           <modal
             name="example-usage-modal"
-            :height="auto"
+            :height="400"
+            :width="650"
```
surchs, line 101: maybe big edge case, but: my poor simple mobile phone has a screen size of 375 x 667 (at least according to dev tools). Does that still work in this case?

surchs (approved):
:cook: 

One slightly edge case question: if the size is fixed: how does that behave across different screen sizes?

PR #270, opened 2023-11-09
[ENH] Fixed the order of modality labels

### LINKED ISSUE #238: Ensure that the modality "labels" are ordered the same way.
Ensure that the "labels" of the the different modality on the card of each dataset are always ordered the same way.

Maybe it is my OCD but having:

- T1w, T2W, flow

and then
- flow, T2W

is a bit disturbing in terms of UX.

rmanaem:
Closes #238 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultsContainer.vue
+++ b/components/ResultsContainer.vue
@@ -75,7 +75,7 @@
           :dataset-uuid="res.dataset_uuid"
           :dataset-name="res.dataset_name"
           :num-matching-subjects="res.num_matching_subjects"
-          :image-modals="res.image_modals"
+          :image-modals="res.image_modals.sort()"
           :is-checked="selectAll || downloads.includes(res.dataset_uuid)"
           :data-cy="res.dataset_uuid"
           @update-downloads="updateDownloads"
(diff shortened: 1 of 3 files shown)


components/ResultsContainer.vue
```diff
@@ -75,7 +75,7 @@
           :dataset-uuid="res.dataset_uuid"
           :dataset-name="res.dataset_name"
           :num-matching-subjects="res.num_matching_subjects"
-          :image-modals="res.image_modals"
+          :image-modals="res.image_modals.sort()"
```
surchs: could we have a test for this that breaks before the PR? If too tricky: please make an issue to add.

surchs (approved):
looks good, but test would be nice so it doesn't sneak back

surchs:
:cook:

components/ResultsContainer.vue
rmanaem: Done

PR #295, opened 2023-11-21
[ENH] Implemented new field query to allow for querying specific nodes

### LINKED ISSUE #267: Allow for querying specifc nodes
For OpenNeuro we need to have a way to only query OpenNeuro data. But when we do that, users should also see that there are other node APIs they could also query by making a manual selection.

TODO:
- [x] Implement a UI element (e.g., dropdown) and required logic to allow users to query a specific node from the available nodes in the federation api
- [x] Have the label be "Neurobagel graph database" or similar

### LINKED ISSUE #283: Add new environment variable to indicate type of API being requested
(Maybe a boolean?) Where the options would reflect either a node API or a federation API.

rmanaem:
- Closes #267
- Closes #283 

- Added `isFederationAPI` env var
  - if `isFederationAPI` is `true`, `index.vue` makes a request to the `get nodes` endpoint and passes them as props along with the value of `isFederationAPI` to the `QueryForm`
- Implemented a UI element for selecting nodes
  - only rendered if the `isFederationAPI` is `true`
  - Receives its options as `node` prop from parent `index.vue` 
- Updated `QueryForm` component test

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -10,6 +10,15 @@
     <b-row>
       <b-form @submit.prevent="validateQueryForm">
         <b-form-row class="row">
+          <categorical-field
+            v-if="isFederationAPI"
+            name="Neurobagel graph database"
+            data-cy="nodes-field"
+            :options="['All', ...Object.keys(nodes)]"
+            multiple="true"
+            default-selected="All"
+            @update-categorical-field="updateField"
+          />
           <b-form-group class="col-md-6">
             <continuous-field
               name="Min Age"
@@ -102,6 +111,8 @@ export default {
   emits: ['update-response'],
   data() {
     return {
+      nodes: {},
+      selectedNodes: [],
       minAge: null,
       maxAge: null,
       sex: null,
@@ -111,7 +122,18 @@ export default {
       assessment: null,
       modality: null,
       isFetching: false,
+      isFederationAPI: true,
     };
+  },
+  async mounted() {
+    this.isFederationAPI = this.$config.isFederationAPI;
+    if (this.isFederationAPI) {
+      const response = await this.$axios.get(`${this.$config.apiQueryURL}nodes/`);
+      this.nodes = response.data.reduce((acc, node) => ({
+        ...acc,
+        [node.NodeName]: node.ApiURL,
+      }), {});
+    }
   },
   methods: {
     updateField(name, input) {
@@ -137,6 +159,11 @@ export default {
         case 'Imaging modality':
           this.modality = this.categoricalOptions[name][input];
           break;
+        case 'Neurobagel graph database':
+          if (!input.includes('All')) {
+            this.selectedNodes = input.map((el) => this.nodes[el]);
+          }
+          break;
         default:
           break;
       }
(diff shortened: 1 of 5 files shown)


components/QueryForm.vue
```diff
@@ -98,10 +107,19 @@ export default {
       type: Object,
       required: true,
     },
+    isFederationAPI: {
+      type: Boolean,
+      default: true,
```
surchs, line 112: I think this should be false, mainly to not change the old default (of this not being there).

components/QueryForm.vue
```diff
@@ -10,6 +10,15 @@
     <b-row>
       <b-form @submit.prevent="validateQueryForm">
         <b-form-row class="row">
+          <categorical-field
+            v-if="isFederationAPI"
+            name="Neurobagel graph database"
```
```suggestion
            name="Select neurobagel graphs to query"
```

components/QueryForm.vue
```diff
@@ -137,6 +155,11 @@ export default {
         case 'Imaging modality':
           this.modality = this.categoricalOptions[name][input];
           break;
+        case 'Neurobagel graph database':
+          if (!input.includes('All')) {
+            this.selectedNodes = input.map((el) => this.nodes[el]);
```
```suggestion
            this.selectedNodeURLs = input.map((node_name) => this.nodes[node_name]);
```
for clarity, and if I understand right.

pages/index.vue
```diff
@@ -106,6 +110,17 @@ export default {
       },
     };
   },
+  async mounted() {
+    if (this.$config.isFederationAPI) {
+      this.isFederationAPI = this.$config.isFederationAPI;
+
+      const response = await this.$axios.get(`${this.$config.apiQueryURL}nodes/`);
+      this.nodes = response.data.reduce((acc, node) => ({
```
surchs:
```suggestion
      this.nodes = response.data.reduce((temp_node_array, node) => ({
```
or something less cryptic

pages/index.vue
```diff
@@ -106,6 +110,17 @@ export default {
...
+        ...acc,
+        [node.NodeName]: node.ApiURL,
```
surchs: OK so `...acc` is my deconstructed, accumulated object here of the form `{ nodeName: nodeURL, ...}`. And now I add another key:value pair here (having to use the `[ ]` syntax. 

This is me not being familiar with either reduce or object deconstruction here - and that's definitely something we should expect to be known. So I wouldn't suggest you add comments. But you can make people's lives easier by naming the variables a little more informatively.

surchs (approved):
Thanks @rmanaem, this looks great! I left a couple of clarity / understanding comments and tried out the PR on the staging deploy (so haven't tested the requests myself).

Everything looks very nice. Take a look at the comments and then :cook:

components/QueryForm.vue
rmanaem: I set the default to `true` because I recall when we discussed this last week, we decided from now on the default assumption is that the query tool will be talking to a federation API (in most cases).

components/QueryForm.vue
rmanaem: How about `Neurobagel graph` to align with the existing field labels?

components/QueryForm.vue
surchs: OK, good point. Let's go with that then

pages/index.vue
rmanaem: Yea, the `reduce` syntax was unfamiliar to me as well.
But I learned `[ ]` is used to dynamically set the property key of an object.

PR #303, opened 2023-11-24
[ENH] Encode selected query nodes as query parameters in URL

### LINKED ISSUE #302: there is a URL / link that leads to a query tool that only searches OpenNeuro
We want

- [x] when I select a node from the dropdown list, my browser URL changes, e.g. q.nb.org/?nodes=openneuro
- [x] when I navigate to such a changed URL, the app state reflects the selection (i.e. looks the same)
- [x] if I enter a URL with some node selected and after the query tool has finished "fetching" all the nodes it is aware of it doesn't know about the node in my URL, it falls back to defaults (e.g. "ALL" nodes selected).

This is not very difficult. 

To put a component variable into the URL:

```js
this.$router.push({ query: { node: "world" } });
```
where **node** is the query parameter and `"world"` is the value -> my.app/?node=world

To read a URL query parameter from the URL into a component variable:

```js
computed: {
    node() {
      return this.$route.query.node;
    },
  },
```
So from the above URL we would get `node="world"` again

What might be trickier is the logic.

So 

1. URL is defined for e.g. `q.nb.org/?node=openneuro`
2. App loads and has to async fetch the list of nodes

What should happen
- until the app gets the list of nodes
- if the URL defined node is not in the list of nodes

surchs:
- Closes #302

- at the beginning of the lifecycle, right after reading node-list from f-API, parse the current browser URL for `node` query parameters
- if `node` query parameters exist, then see if they describe available nodes
- if they don't, default to the "All" node
- if they do, set these nodes to selected immediately
- make all of this reactive (i.e. if user selects nodes in dropdown, URL changes and vice versa)
- change how nodes are represented internally to be an array (of exactly the same shape as the response object from the f-API/nodes endpoint)
- work around the "All" node not having a defined ApiURL

**Note**: due to running out of brain juice this PR does not have any tests for this new functionality. That's a problem and we should fix it either before we merge or directly after. The challenging bit is going to be intercepting the URLs and environment variables.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -129,7 +129,6 @@ export default {
   emits: ['update-response', 'select-nodes'],
   data() {
     return {
-      selectedNodeNames: [],
       minAge: null,
       maxAge: null,
       sex: null,
@@ -141,15 +140,7 @@ export default {
       isFetching: false,
     };
   },
-  computed: {
-    selectedNodeURLs() {
-      return this.selectedNodeNames.map((nodeName) => this.availableNodes[nodeName]);
-    },
-  },
   methods: {
-    nodeWasSelected(event) {
-      this.$emit('selectNodes', event);
-    },
     updateField(name, input) {
       switch (name) {
         case 'Min Age':
@@ -197,9 +188,9 @@ export default {
     async submitQuery() {
       this.isFetching = true;
       let url = `${this.$config.apiQueryURL}query/?`;
-      if (this.isFederationAPI && this.selectedNodeURLs.length > 0) {
+      if (this.isFederationAPI && this.selectedNodes.length > 0) {
         this.selectedNodeURLs.forEach((node) => {
-          url += `&node_url=${node}`;
+          url += `&node_url=${node.ApiURL}`;
         });
       }
       if (this.minAge) {
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -11,9 +11,9 @@
       <b-form @submit.prevent="validateQueryForm">
         <b-form-row class="row">
           <v-select
-            v-model="selected_nodes"
+            v-model="selectedNodeNames"
             data-cy="node-field"
-            :options="['All', ...Object.keys(neurobagel_nodes)]"
+            :options="['All', ...Object.keys(availableNodes)]"
             :multiple=true
           />
           <b-form-group class="col-md-6">
@@ -116,8 +116,8 @@ export default {
   emits: ['update-response'],
   data() {
     return {
-      neurobagel_nodes: [],
-      selected_nodes: [],
+      availableNodes: [],
+      selectedNodeNames: [],
       minAge: null,
       maxAge: null,
       sex: null,
@@ -131,7 +131,7 @@ export default {
   },
   async fetch() {
     const response = await this.$axios.get(`${this.$config.apiQueryURL}nodes/`);
-    this.neurobagel_nodes = response.data.reduce((tempNodeArray, node) => ({
+    this.availableNodes = response.data.reduce((tempNodeArray, node) => ({
       ...tempNodeArray,
       [node.NodeName]: node.ApiURL,
     }), {});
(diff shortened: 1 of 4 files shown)


pages/index.vue
```diff
@@ -110,23 +112,82 @@ export default {
...
+      return this.$config.isFederationAPI === undefined ? true : this.$config.isFederationAPI;
+    },
+  },
+  watch: {
+    selectedNodes(newNode) {
+      if (this.isFederationAPI && newNode.length > 0) {
+        this.$router.push({
+          path: this.$route.path,
+          query: { ...this.$route.query, node: newNode.map((node) => node.NodeName) },
+        });
```
rmanaem: I think this doesn't take the case where all options have been cleared from the `v-select` into account when it updates the router.

pages/index.vue
```diff
@@ -110,23 +112,82 @@ export default {
       },
     };
   },
-  async mounted() {
-    if (this.$config.isFederationAPI) {
-      this.isFederationAPI = this.$config.isFederationAPI;
-
+  async fetch() {
+    if (this.isFederationAPI) {
+      // TODO: write a test for all these fancy things
```
rmanaem, line 117: I see you added a TODO for tests. Is that because the changes in this PR are time sensitive?

rmanaem: Or figuring out how to test these changes wouldn't be worthwhile at the moment?

components/QueryForm.vue
```diff
@@ -184,9 +188,11 @@ export default {
     async submitQuery() {
       this.isFetching = true;
       let url = `${this.$config.apiQueryURL}query/?`;
-      if (this.selectedNodeURLs.length > 0) {
-        this.selectedNodeURLs.forEach((node) => {
-          url += `&node_url=${node}`;
+      if (this.isFederationAPI && this.selectedNodes.length > 0) {
```
rmanaem: Not sure if we need to check for `isFederationAPI` as when its set to false, the `selectedNodes` would never be updated and its default value would render this condition false.

rmanaem (changes requested):
Thanks for the PR @surchs! 
I left a couple comments, please take a look. 
Aside from those, I noticed an issue with the `isFederationAPI` when its explicitly set to false in the `.env`.
The value is correctly determined to be false in the `index.vue` but appears as true in the `QueryForm` therefore the `div` containing the Neurobagel graph dropdown gets rendered. Would you mind looking into this?

rmanaem:
> Aside from those, I noticed an issue with the isFederationAPI when its explicitly set to false in the .env.
The value is correctly determined to be false in the index.vue but appears as true in the QueryForm therefore the div containing the Neurobagel graph dropdown gets rendered. Would you mind looking into this?

It seems this stems from not following the vue prop naming conventions so changing the prop to `isFederationApi` should fix it.

pages/index.vue
surchs: good point!

pages/index.vue
surchs: Tests we should write:
component
- query form component test for select-nodes
  - emit new event
  - render props nicely
e2e
- the new URL query pushing thing
- setting isFederationApi to false should actually work as intended
-

surchs:
> > Aside from those, I noticed an issue with the isFederationAPI when its explicitly set to false in the .env.
> > The value is correctly determined to be false in the index.vue but appears as true in the QueryForm therefore the div containing the Neurobagel graph dropdown gets rendered. Would you mind looking into this?
> 
> It seems this stems from not following the vue prop naming conventions so changing the prop to `isFederationApi` should fix it.

For posterity / clarity: what's happening here is:
- https://eslint.vuejs.org/rules/attribute-hyphenation.html requires that props are passed with hyphenation
- we therefore pass `isFederatedAPI` to the `QueryForm.vue` component as `is-federated-api`
- when the `QueryForm.vue` component retrieves the prop, it unpacks it as `isFederatedApi`, presumably because the fact that all last three characters were capitalized is lost and it just defaults to camelCase
- we don't use `isFederatedApi` in the `QueryForm.vue` component, but we do use `isFederatedAPI` - which is now unset
- because `isFederatedAPI` also has a default value, and because this default value is `true`, we have now created a situation where `isFederatedAPI` is always going to be true inside the component (because it will never receive a prop value)

Therefore the only solution is to rename `isFederatedAPI` -> `isFederatedApi` **inside** of `QueryForm.vue`. Also renaming it in `index.vue` is entirely optional, but probably a better idea for consistency's sake.

surchs:
@rmanaem about testing `isFederationAPI` being correctly parsed from the environment: I think this is trickier than just using `Cypress.env` and I'd rather address it in a future PR than holding this up until we understand it.

e.g. this (somewhat reasonable looking test) does not work:

```js
describe('API request', () => {
  it('Parses the isFederationApi environment variable correctly', () => {
    cy.visit('/');
    Cypress.env('isFederationApi', false);
    cy.get('[data-cy="node-field"]').should('not.exist');
    Cypress.env('isFederationApi', true);
    cy.get('[data-cy="node-field"]').should('exist');
  });
});
```

This could have to do with the fact that although `isFederationApi` is a computed property:

```js
computed: {
    isFederationApi() {
      return this.$config.isFederationAPI === undefined ? true : this.$config.isFederationAPI;
    },
  },
```
it has no way of detecting a change in the environment and only is read / updated once. Or maybe I'm not setting the `.env` correctly. Either way, we should investigate after this is merged!

rmanaem (approved):
:cook:

PR #307, opened 2023-11-29
[FIX] Implemented safeguard for falsy modalities

### LINKED ISSUE #305: [BUG] Safeguard against `undefined` modalities

### Expected Behavior

The app should behave the same in both environments.

### Current Behavior

When running the app in `dev` mode on the `main` branch the app seems to break and respond with a rather ambiguous error once the `Submit Query` button is clicked. However, the app runs just fine when running in production and was built by running `npm run generate` and `npm run start`

### Error message

Screenshot from 2023-11-28 17-34-01

image

### Environment

_No response_

### How to reproduce

`npm run dev` on the `main` branch, then run an empty query 
vs
`npm run generate` and `npm run start`, then run an empty query

### Anything else?

_No response_

rmanaem:
- Closes #305 

- Filtered modalities for falsy values
- Added a test in the `ResultCard` component with falsy modalities 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/ResultCard.vue
+++ b/components/ResultCard.vue
@@ -64,7 +64,7 @@
                   class="float-right"
                 >
                   <b-button
-                    v-for="modality in sortedModalities"
+                    v-for="modality in sortedFilteredModalities"
                     :key="modality"
                     type="button"
                     class="btn card-modality"
@@ -149,10 +149,10 @@ export default {
     };
   },
   computed: {
-    sortedModalities() {
+    sortedFilteredModalities() {
       // We have to do this as the linter screams at us for mutating the prop
-      const sortedModalities = this.imageModals;
-      return sortedModalities.sort();
+      const modalities = this.imageModals;
+      return modalities.filter(Boolean).sort();
     },
   },
   methods: {
--- a/components/ResultCard.vue
+++ b/components/ResultCard.vue
@@ -152,7 +152,8 @@ export default {
     sortedFilteredModalities() {
       // We have to do this as the linter screams at us for mutating the prop
       const modalities = this.imageModals;
-      return modalities.filter(Boolean).sort();
+      // Filter out null modalities as the API may return null for some modalities
+      return modalities.filter((modality) => modality !== null).sort();
     },
   },
   methods: {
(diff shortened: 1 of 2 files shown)


cypress/component/ResultCard.cy.js
```diff
@@ -46,4 +46,10 @@ describe('Result card', () => {
     cy.get('[data-cy="card-http://neurobagel.org/vocab/cool-dataset-expander"]').contains('...less').click();
     cy.get('[data-cy="card-http://neurobagel.org/vocab/cool-dataset-dataset"]').should('not.contain', 'truncated');
   });
+  it('Filters falsy modalities', () => {
+    props.imageModals = [null, undefined];
+    cy.mount(ResultCard, {
+      propsData: props,
+    });
+  });
```
surchs: nice, but let's make an assert too so this can fail. e.g. we could add a non-falsy modality and assert that it renders.

components/ResultCard.vue
```diff
@@ -149,10 +149,10 @@ export default {
     };
   },
   computed: {
-    sortedModalities() {
+    sortedFilteredModalities() {
       // We have to do this as the linter screams at us for mutating the prop
-      const sortedModalities = this.imageModals;
-      return sortedModalities.sort();
+      const modalities = this.imageModals;
+      return modalities.filter(Boolean).sort();
```
surchs: I think this is too clever for the level of readability we're aiming for. Another issue is that a bunch of values other than `null` [are falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), and we don't know yet if they are all going to be irrelevant in the future.

I would suggest we write the long form and just filter for `null` for now. When we encounter `undefined` we can either add that too or think again. In either case I would also add a comment to explain that sometimes the API can return `null` for modalities and that's why we remove them here.

surchs (commented):
thanks for the quick fix @rmanaem. two changes please,and then good

surchs (approved):
:tada: 

:cook:

PR #308, opened 2023-11-29
[ENH] Switched to requesting available terms for diagnosis and assessment tool from API

### LINKED ISSUE #261: Switch to requesting available terms for queryable variables from linked API(s)
Now that https://github.com/neurobagel/federation-api/issues/12 is done, the f-API has an endpoint to get available terms from all available n-APIs.

We should now

- [x] Fetch the available terms for diagnosis and assessment tool from the new f-API endpoint
- [x] Decide at what stage of the lifecycle we should make the request
- [x] Remove the old hardcoded diagnosis and assessment options.
- [x] Provide a nice error message (e.g. https://bootstrap-vue.org/docs/components/toast) if either or both of these requests fail and we get zero terms back from the f-API

rmanaem:
- Closes #261 

- Removed the hardcoded options for diagnosis and assessment tool
- Switched to requesting available terms for diagnosis and assessment tool from API
- Sorted categorical options since the assessment tool terms included a mix of lowercase and uppercase labels
- Implemented logic to display a message when the diagnosis and assessment tool options are not retrieved

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/CategoricalField.vue
+++ b/components/CategoricalField.vue
@@ -12,7 +12,7 @@
     <v-select
       v-model="selected"
       :data-cy="`${name}-select`"
-      :options="options"
+      :options="sortedOptions"
       :multiple="multiple"
       :disabled="disabled"
       @input="updateField"
@@ -49,6 +49,12 @@ export default {
     return {
       selected: this.defaultSelected,
     };
+  },
+  computed: {
+    sortedOptions() {
+      const { options } = this;
+      return options.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
+    },
   },
   methods: {
     updateField() {
--- a/components/CategoricalField.vue
+++ b/components/CategoricalField.vue
@@ -52,6 +52,7 @@ export default {
   },
   computed: {
     sortedOptions() {
+      // We have to do this to avoid mutating the prop
       const { options } = this;
       return options.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
     },
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -52,7 +52,7 @@
         <categorical-field
           name="Diagnosis"
           data-cy="diagnosis-field"
-          :default-selected="Object.keys(categoricalOptions.Diagnosis)[0]"
+          default-selected="All"
           :options="Object.keys(categoricalOptions.Diagnosis)"
           :disabled="is_control"
           @update-categorical-field="updateField"
@@ -79,7 +79,7 @@
         <categorical-field
           name="Assessment tool"
           data-cy="assessment-tool-field"
-          :default-selected="Object.keys(categoricalOptions['Assessment tool'])[0]"
+          default-selected="All"
           :options="Object.keys(categoricalOptions['Assessment tool'])"
           @update-categorical-field="updateField"
         />
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -20,7 +20,7 @@
               :value="selectedNodes"
               data-cy="node-field"
               :options="availableNodes"
-              :multiple=true
+              :multiple="true"
               label="NodeName"
               @input="$emit('selectNodes', $event)"
             />
@@ -139,6 +139,14 @@ export default {
       modality: null,
       isFetching: false,
     };
+  },
+  mounted() {
+    if (Object.keys(this.categoricalOptions.Diagnosis).length <= 1) {
+      this.displayToast('Failed to retrieve diagnosis options');
+    }
+    if (Object.keys(this.categoricalOptions['Assessment tool']).length <= 1) {
+      this.displayToast('Failed to retrieve assessment tool options');
+    }
   },
   methods: {
     updateField(name, input) {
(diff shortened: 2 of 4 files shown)


components/CategoricalField.vue
```diff
@@ -50,6 +50,12 @@ export default {
       selected: this.defaultSelected,
     };
   },
+  computed: {
+    sortedOptions() {
+      const { options } = this;
```
surchs, line 56: we need this because `.sort()` sorts in place and that would break the "do not mutate props" rule, yes? I think that's borderline self-evident. So maybe a "why we do this" comment could help here.

components/CategoricalField.vue
```diff
@@ -50,6 +50,12 @@ export default {
...
+      return options.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
```
surchs, line 57: can we give this some better names? `a` -> `label`, `b` -> `termURL` or something like that?

components/QueryForm.vue
```diff
@@ -20,7 +20,7 @@
               :value="selectedNodes"
               data-cy="node-field"
               :options="availableNodes"
-              :multiple=true
+              :multiple="true"
```
surchs: I think this is a bit of a linter-tug-of-war. iirc eslint wants this quoted as string, but v-select wants it as boolean. It seems we can just change to 
```suggestion
              multiple
```
based on https://vue-select.org/guide/values.html#single-multiple, maybe try that.

components/QueryForm.vue
```diff
@@ -52,7 +52,7 @@
         <categorical-field
           name="Diagnosis"
           data-cy="diagnosis-field"
-          :default-selected="Object.keys(categoricalOptions.Diagnosis)[0]"
+          default-selected="All"
```
surchs, line 55: I like this, and I know we call the option "All" everywhere, but I wonder if `"Any"` would be more intuitive. Because we're not really searching for one subject who has all of the diagnoses, but rather any. Leaving as note for future issue.

cypress/component/QueryForm.cy.js
```diff
@@ -141,4 +141,14 @@ describe('Query form', () => {
...
   });
+  it('Displays a toast when diagnosis and assessment tool options are not retrieved', () => {
+    props.categoricalOptions.Diagnosis = { All: null };
+    props.categoricalOptions['Assessment tool'] = { All: null };
+    cy.mount(QueryForm, {
+      stubs,
+      propsData: props,
+    });
+    cy.contains('#b-toaster-top-right', 'Failed to retrieve diagnosis options');
+    cy.contains('#b-toaster-top-right', 'Failed to retrieve assessment tool options');
```
surchs, line 206: interesting. Why don't you use `cy.get` here first?

pages/index.vue
```diff
@@ -33,73 +33,8 @@ export default {
...
-          'Parkinson\'s disease': 'snomed:49049000',
-          'Progressive supranuclear palsy': 'snomed:192976002',
-          'Separation anxiety': 'snomed:126943008',
-          Schizophrenia: 'snomed:58214004',
-          'Specific spelling disorder': 'snomed:268738002',
-          'Traumatic brain injury': 'snomed:127295002',
-          'Visual impairment': 'snomed:397540003',
         },
         'Assessment tool': {
-          All: null,
```
surchs, line 59: :tada:

pages/index.vue
```diff
@@ -113,6 +48,22 @@ export default {
...
+    this.categoricalOptions.Diagnosis = diagnosisOptions;
+
+    const assessmentResponse = await this.$axios.get(`${this.$config.apiQueryURL}attributes/nb:Assessment`);
+    const assessmentOptions = assessmentResponse.data['nb:Assessment'].reduce((tempArray, assessment) => ({
+      ...tempArray,
+      [assessment.Label]: assessment.TermURL,
+    }), {});
+    assessmentOptions.All = null;
+    this.categoricalOptions['Assessment tool'] = assessmentOptions;
+
```
surchs, line 66: :tada:

surchs (approved):
great stuff, thanks @rmanaem :tada:!

Two :cherries: readability things, and then :cook:

surchs:
oh, and please check that all is still happy and running after resolving the merge conflicts

components/QueryForm.vue
rmanaem: I like that.

components/QueryForm.vue
rmanaem: Ah yes I didn't notice it finally made its way into a commit.

components/CategoricalField.vue
rmanaem: We could do `label1` and `label2` since `a` and `b` are two elements from the options (labels) array.

cypress/component/QueryForm.cy.js
rmanaem: You can use `cy.contains` with a selector to find an element containing a specific text instead of chaining it off `cy.get`.

PR #309, opened 2023-11-30
[ENH] add trailing slash to queryURL if none was added by user

### LINKED ISSUE #212: Ensure API_QUERY_URL terminates with slash
- Document that the API query URL has to end in forward slash
- Put some sort of check into the query tool for when there is no tailing slash - and then add one

otherwise the request can fail if a port is also specified. E.g.

`API_QUERY_URL=http://localhost:8000` fails

_Originally posted by @surchs in https://github.com/neurobagel/api/pull/154#discussion_r1316470086_
            

Tasks

- [x] if URL without slash is provided -> add a slash
- [x] if URL with slash is provided -> do not add slash
- [x] make test to ensure that's true

surchs:
- Closes #212 

- new test: the only good way to check if the internal parsing of apiQueryURL has worked is to first stub and then spy the axios.get call. So that's what I'm doing here
- I also taught `.eslint` to ignore the `dist/` dir because dist doesn't follow our styleguide and that makes eslint *very* upset

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/component/QueryForm.cy.js
+++ b/cypress/component/QueryForm.cy.js
@@ -141,4 +141,58 @@ describe('Query form', () => {
 
     cy.get('[data-cy="node-field"]').should('contain', 'anotherNode');
   });
+
+  it('Appends a slash to the apiQueryURL if it does not end with a slash', () => {
+    const getStub = cy.stub().resolves({ data: 'mock response' });
+
+    cy.mount(QueryForm, {
+      stubs,
+      propsData: {
+        selectedNodes: [
+          {
+            NodeName: 'anotherNode',
+            NodeURL: 'https://anotherNode.org',
+          },
+        ],
+        ...props,
+      },
+      mocks: {
+        $config: {
+          apiQueryURL: 'http://my.site.org',
+        },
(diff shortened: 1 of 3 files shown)


cypress/component/QueryForm.cy.js
```diff
@@ -141,4 +141,58 @@ describe('Query form', () => {
...
+            NodeName: 'anotherNode',
+            NodeURL: 'https://anotherNode.org',
+          },
+        ],
+        ...props,
+      },
+      mocks: {
+        $config: {
+          apiQueryURL: 'http://my.site.org',
+        },
```
rmanaem, line 162: Is this how we can override/mock env vars in component tests?

rmanaem (approved):
:cook:

cypress/component/QueryForm.cy.js
surchs: Not the env var directly but the global config where you're reading the env var and from where we're accessing it with this.$config

PR #310, opened 2023-11-30
[ENH] Added tooltip for healthy control checkbox

### LINKED ISSUE #163: Diagnosis checkbox unclear
It is unclear what the diagnosis/healthy control checkbox does, besides gray out the diagnosis dropdown. I suggest changing the wording of the text by the box to better indicate its function.

rmanaem:
- Closes #163 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -20,7 +20,7 @@
               :value="selectedNodes"
               data-cy="node-field"
               :options="availableNodes"
-              :multiple=true
+              :multiple="true"
               label="NodeName"
               @input="$emit('selectNodes', $event)"
             />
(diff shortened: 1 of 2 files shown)


components/QueryForm.vue
```diff
@@ -20,7 +20,7 @@
               :value="selectedNodes"
               data-cy="node-field"
               :options="availableNodes"
-              :multiple=true
+              :multiple="true"
```
surchs: same as in the other PR, I think you can just use
```suggestion
              multiple
```
and avoid the linter war. (but do lmk if that's not true)

components/QueryForm.vue
```diff
@@ -59,13 +59,23 @@
...
             <b-form-group class="col-md-7">
-              <input
-                v-model="is_control"
-                data-cy="healthy-control-checkbox"
-                class="form-check-input"
-                type="checkbox"
+              <div
+                v-b-tooltip.hover.right="'Checking the healthy control checkbox \
+              disables the diagnosis field as it is assumed participants \
+              can either be healthy or have diagnosis'"
```
```suggestion
                v-b-tooltip.hover.right="'Checking the healthy control checkbox \
              disables the diagnosis field because we do not support queries
              for healthy controls with secondary diagnoses'"
```

cypress/component/QueryForm.cy.js
```diff
@@ -141,4 +141,12 @@ describe('Query form', () => {
 
     cy.get('[data-cy="node-field"]').should('contain', 'anotherNode');
   });
+  it.only('Displays the healthy control tooltip', () => {
```
surchs:
```suggestion
  it('Displays the healthy control tooltip', () => {
```
debug leftover?

surchs (approved):
Looks good @rmanaem, one fix for the tooltip, then good to go :cook:

components/QueryForm.vue
rmanaem: You're right that works, just gotta merge `main`.

cypress/component/QueryForm.cy.js
rmanaem: Yea, good catch!

PR #311, opened 2023-12-01
[FIX] Fixed the URL parameter resetting in GItHub pages deployed app

rmanaem:
According to Nuxt 2 docs https://v2.nuxt.com/docs/components-glossary/fetch/#nuxt--212, fetch hook can be used to get asynchronous data therefore we'll keep all logic related to fetching asynchronous data in this hook. Since we are currently deploying the app on Github Pages and using static site generation, fetch hook is only called during the generation phase i.e., `npm run generate` and thus any logic related to updating the route would need to be moved to the mounted hook which is run when the component is inserted into the DOM including when the app is initialized or reinitialized.

## Checklist

- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -47,6 +47,12 @@ export default {
       },
     };
   },
+  // According to Nuxt 2 docs https://v2.nuxt.com/docs/components-glossary/fetch/#nuxt--212,
+  // fetch hook can be used to get asynchronous data therefore we'll keep all logic related
+  // to fetching asynchronous data in this hook
+  // As we are currently deploying the app on Github Pages and using static site generation,
+  // fetch hook is only called during generation phase i.e., npm run generate and thus
+  // any logic related to updating the route would need to be moved to mounted hook
   async fetch() {
     const diagnosisResponse = await this.$axios.get(`${this.$config.apiQueryURL}attributes/nb:Diagnosis`);
     const diagnosisOptions = diagnosisResponse.data['nb:Diagnosis'].reduce((tempArray, diagnosis) => ({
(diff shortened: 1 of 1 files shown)


rmanaem:
For the reviewer: ignore the netlify deploy preview and check the GitHub pages deployment.

pages/index.vue
```diff
@@ -138,6 +98,47 @@ export default {
       }
     },
   },
+  mounted() {
+    // The first time we load the app, we will also check the URL
+    // for valid query parameters that refer to selected nodes.
+    // If we find any valid query parameters in the URL, then we
+    // initialize the apps with these nodes selected!
```
surchs:
```suggestion
    // Once the app is ready to render, we will also check the URL
    // for valid query parameters that refer to selected nodes.
    // If we find any valid query parameters in the URL, then we
    // initialize the apps with these nodes selected!
    // This has to happen at the onMounted lifecycle event
    // because the URL is not available earlier
```

surchs: Or something similar to make it clear what you have found out about the lifecycle events in static mode.

surchs (approved):
Thanks a lot @rmanaem for investigating and fixing.

Before we merge this, we have to be clear among ourselves why this broke and why it works now. I.e. when you moved from `fetch` to `mounted` for everything. the vocab-getting broke, right? So the fact that we are now using both has to be clearly communicated. 

Please:
1. Add a very brief comment at the top of both the `fetch` and `mounted` hooks to explain why the specific content goes there and not in the other section
2. Comment on the PR to summarize any more detailed info you have learned while fixing this.

Then we can merge :cook:

PR #315, opened 2023-12-05
[ENH] Added an alert for `OpenNeuro` node limitations

### LINKED ISSUE #294: Add OpenNeuro specific note on limitations
When I select "OpenNeuro" as a query node (#267) a dedicated visual element appears prominently on the query tool and 

- [x] explains that OpenNeuro search is being tested / developed
- [x] explains that we don't cover all of the OpenNeuro datasets
- [x] provides a link to a site where I can learn more, and help annotate more data (https://github.com/neurobagel/planning/issues/77)

rmanaem:
- Closes #294 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/e2e/Alert.cy.js
+++ b/cypress/e2e/Alert.cy.js
@@ -1,9 +1,9 @@
 describe('Alert', () => {
   it('Correctly displays and dismisses the alert', () => {
-    cy.visit('/');
-    cy.get('[data-cy="alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
     cy.visit('/?node=Parkinson\'s Progression Markers Initiative');
     cy.get('[data-cy="alert"]').should('not.exist');
+    cy.visit('/');
+    cy.get('[data-cy="alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
     cy.get('[data-cy="node-field"]').type('OpenNeuro{enter}');
     cy.get('[data-cy="alert"]').should('be.visible');
     cy.get('[data-cy="dismiss-alert"]').click();
--- a/cypress/e2e/Alert.cy.js
+++ b/cypress/e2e/Alert.cy.js
@@ -1,12 +1,12 @@
 describe('Alert', () => {
   it('Correctly displays and dismisses the alert', () => {
-    cy.visit('/');
-    cy.get('[data-cy="alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
+    cy.visit('/?node=All');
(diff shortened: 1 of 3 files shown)


pages/index.vue
```diff
@@ -90,6 +114,9 @@ export default {
       // So to keep things consistent, we call it isFederationApi here as well.
       return this.$config.isFederationAPI === undefined ? true : this.$config.isFederationAPI;
     },
+    queryingOpenNeuro() {
+      return this.selectedNodes.some((node) => node.NodeName === 'OpenNeuro' || node.NodeName === 'All') && !this.alertDismissed;
+    },
```
surchs, line 119: neat

cypress/e2e/Alert.cy.js
```diff
@@ -0,0 +1,12 @@
+describe('Alert', () => {
+  it('Correctly displays and dismisses the alert', () => {
+    cy.visit('/');
```
surchs: that's cool, but maybe a bit too smart. I'd say either visit node=All directly or even OpenNeuro. Otherwise we're both implicitly testing the node-reset function (which is unrelated) and we have to remember that a blank URL will also query OpenNeuro

cypress/e2e/Alert.cy.js
```diff
@@ -0,0 +1,12 @@
...
+    cy.get('[data-cy="alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
+    cy.visit('/?node=Parkinson\'s Progression Markers Initiative');
+    cy.get('[data-cy="alert"]').should('not.exist');
+    cy.get('[data-cy="node-field"]').type('OpenNeuro{enter}');
```
surchs, line 7: I don't mind either way of setting the nodes. But is there a reason you're switching it up between URL-setting and UI-setting?

cypress/e2e/Alert.cy.js
```diff
@@ -0,0 +1,12 @@
+describe('Alert', () => {
+  it('Correctly displays and dismisses the alert', () => {
+    cy.visit('/');
+    cy.get('[data-cy="alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
```
surchs: I think it makes sense to assert part of the text in the alert. But maybe we could also give the alert an OpenNeuro specific data-cy, so we can just assert that the right alert is shown (even though there is only one for now_

surchs (approved):
very nice! one :cherries: for the test, then :cook:

cypress/e2e/Alert.cy.js
rmanaem: No just going straight with url setting here to avoid repetition since we have to select something other than `OpenNeuro` so we can then remove `All`.

PR #317, opened 2023-12-06
[FIX] Refactored `submitQuery` to resolve the query url construction bug

### LINKED ISSUE #316: [BUG] minimum age gets melted into node API in query

### Expected Behavior

I can run a query for a single node and define a minimum age

### Current Behavior

if I set sex, all is well: `https://federate.neurobagel.org/query/?&node_url=https://api.neurobagel.org/&sex=snomed:248153007`

if I set minimum age, this happens: `https://federate.neurobagel.org/query/?&node_url=https://api.neurobagel.org/min_age=1` 

### Error message

federation API rightly complains that 

`"Unrecognized Neurobagel node URL(s): ['https://api.neurobagel.org/min_age=1/']`

### Environment

- any single or multiple node selection except "All"

### How to reproduce

1. Select any node other than "All"
2. define a minimum age
3. watch it break with "Oops"

### Anything else?

see also https://github.com/OpenNeuroOrg/openneuro/issues/2807

most likely this is our culprit: 
https://github.com/neurobagel/query-tool/blob/dadbee332c593f95f5bab9bc037837f2241a6f76/components/QueryForm.vue#L222-L227

i.e. we do not concatenate `min_age` correctly. 

The reason this doesn't break for the "All" node is that we encode "All" as "no node parameter set":

`https://federate.neurobagel.org/query/?min_age=1`

so in this very specific case, `min_age` actually is the first query parameter and gets parsed correctly. This is most likely a leftover thing from when a query tool only ever talked directly to a node API and the federation API didn't exist yet.

rmanaem:
- Closes #316 

- Refactored `submitQuery` into two methods
  - `constructQueryURL` method which will use `URLSearchParams` to construct
the query param portion of the query URL
  - `submitQuery` will simply call `constructQueryURL` and then make the
axios call
- Updated `APIRequest` test

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -228,7 +228,7 @@ export default {
       queryParams.image_modal = this.modality;
 
       Object.keys(queryParams).forEach((key) => {
-        if (queryParams[key] == null || queryParams[key] === undefined) {
+        if (queryParams[key] === null || queryParams[key] === undefined) {
           delete queryParams[key];
         }
       });
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -227,8 +227,11 @@ export default {
       queryParams.assessment = this.assessment;
       queryParams.image_modal = this.modality;
 
-      Object.keys(queryParams).forEach((key) => queryParams[key] == null
-      && delete queryParams[key]);
+      Object.keys(queryParams).forEach((key) => {
+        if (queryParams[key] == null || queryParams[key] === undefined) {
+          delete queryParams[key];
+        }
+      });
       const searchParams = new URLSearchParams(queryParams).toString();
       return `${this.apiQueryURL}${searchParams}`;
     },
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -209,40 +209,34 @@ export default {
         this.submitQuery();
       }
     },
-    async submitQuery() {
-      this.isFetching = true;
-      let url = this.apiQueryURL;
+    constructQueryURL() {
+      const queryParams = {};
       if (this.isFederationApi && this.selectedNodes.length > 0) {
         this.selectedNodes.forEach((node) => {
           if (node.NodeName !== 'All') {
-            url += `&node_url=${node.ApiURL}`;
+            queryParams.node_url = node.ApiURL;
           }
         });
       }
-      if (this.minAge) {
-        url += `min_age=${this.minAge}`;
-      }
-      if (this.maxAge) {
-        url += `&max_age=${this.maxAge}`;
-      }
-      if (this.sex) {
-        url += `&sex=${this.sex}`;
-      }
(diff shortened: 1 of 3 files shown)


components/QueryForm.vue
```diff
@@ -209,40 +209,34 @@ export default {
...
+      queryParams.max_age = this.maxAge;
+      queryParams.sex = this.sex;
+      queryParams.diagnosis = this.diagnosis && !this.is_control ? this.diagnosis : null;
+      queryParams.is_control = this.is_control || null;
+      queryParams.min_num_sessions = this.min_num_sessions;
+      queryParams.assessment = this.assessment;
+      queryParams.image_modal = this.modality;
+
+      Object.keys(queryParams).forEach((key) => queryParams[key] == null
+      && delete queryParams[key]);
```
surchs: :cherries: 
I think this is one of those cases where a classic for loop would be more readable. 
That said, making a filtered object here rather than deleting in place would probably be the cleanest solution. 

e.g.
```js
Object.fromEntries(
  Object.entries(queryParams)
.filter(([param, value]) => value !== null)
```

components/QueryForm.vue
```diff
@@ -209,40 +209,34 @@ export default {
...
+      queryParams.min_age = this.minAge;
+      queryParams.max_age = this.maxAge;
+      queryParams.sex = this.sex;
+      queryParams.diagnosis = this.diagnosis && !this.is_control ? this.diagnosis : null;
+      queryParams.is_control = this.is_control || null;
+      queryParams.min_num_sessions = this.min_num_sessions;
+      queryParams.assessment = this.assessment;
+      queryParams.image_modal = this.modality;
+
+      Object.keys(queryParams).forEach((key) => queryParams[key] == null
```
```
== null
```
is this a sneaky way to catch `undefined`? I'd say let's be explicit then and list both!

surchs (approved):
thanks a lot for addressing this so quickly @rmanaem! and for finding this much better way of URL parameter construction!

All looks good, but I left a comment on the new URL object filter section, I think we can make that clearer

:cook:

components/QueryForm.vue
rmanaem: Yea but I don't think `undefined` can make its way in there since the defaults are `null` and `v-select` also returns `null` when we remove the selection.

surchs: Ok then let's use strict equal

components/QueryForm.vue
rmanaem: eslint didn't like this implementation because `param` is unused.

components/QueryForm.vue
rmanaem: switched to strict equal and added the check for `undefined`.

PR #335, opened 2023-12-18
[ENH] Updated `QueryForm` layout to clarify query parameters

### LINKED ISSUE #285: Clarify query parameter names
While we are still only handling non-session-level phenotypic information, we should update the names of the query parameters in the UI accordingly so there is no confusion about which timepoint users are querying for.

These include:
"Min/max age" -> "Min/max age at baseline"
"Diagnosis" -> "Diagnosis at baseline"
"Minimum number of sessions" -> "Minimum number of imaging sessions"

- See also https://github.com/neurobagel/bagel-cli/issues/225

TODO
- [x] have on-hover tooltip for each query parameter

rmanaem:
- Closes #285 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/components/QueryForm.vue
+++ b/components/QueryForm.vue
@@ -25,6 +25,12 @@
               @input="$emit('selectNodes', $event)"
             />
           </div>
+          <div>
+            <hr>
+          </div>
+          <b-row>
+            <h6>Phenotypic fields (at baseline)</h6>
+          </b-row>
           <b-form-group class="col-md-6">
             <continuous-field
               name="Min Age"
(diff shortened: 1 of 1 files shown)


components/QueryForm.vue
```diff
@@ -25,6 +25,12 @@
               @input="$emit('selectNodes', $event)"
             />
           </div>
+          <div>
+            <hr>
+          </div>
+          <b-row>
+            <h6>Phenotypic fields (at baseline)</h6>
```
surchs: could you try a higher order heading? It's hard to notice that these are subsection headings - they seem to have the same font weight as the rest of the query fields in the preview: 
image

surchs (approved):
One :cherries: comment @rmanaem: could we make the section headings more visible,

then :cook:

components/QueryForm.vue
rmanaem: Yea I used `h5` first but the Phenotypic one will overflow to the next line

image

rmanaem: @surchs What do you think?

surchs: linebreak is fine, we can always make it prettier!

Issue #328, opened 2023-12-12
Handle case where no diagnosis/assessment exist in the graph


alyssadai:
Currently, the query tool emits some error (warning?) messages about "no diagnosis options" being fetched (and similarly for assessment) if the data in the graph doesn't contain any assessment/diagnosis data.

This however doesn't actually prevent you from querying the data that _is_ in the graph (e.g., with an empty query). We should probably either remove this error message or make it more informative to indicate that it doesn't necessarily mean there is a problem with the database.

PR #343, opened 2023-12-19
[ENH] Configured app to run `fetch` hook on client-side only

### LINKED ISSUE #7: Internal IP address of host vs service name for local federation
In the `local_federation` setup, we use `localhost` for the `API_QUERY_URL` to refer to the federation API running on the host machine, This works fine for the requests made by clicking the `submit query` on the query tool interface from the browser (i.e., client side) but this is an issue for requests made in the `fetch` hook cycle that are made from the query tool container inside the docker network for which `localhost` refers to the query tool container. Not the `host` running the docker compose stack, therefore, requests to retrieve diagnosis, assessments, and nodes fail.

To resolve this
- [x] neurobagel/query-tool#336

See also https://github.com/neurobagel/recipes/issues/5

### LINKED ISSUE #336: Run `fetch` only on client-side
- [x] set `fetchOnServer` to false so the `fetch` only runs on client-side
- [x] Update the DOCKEFILE to use `generate` instead of `build`

rmanaem:
- Closes #336 
- Closes https://github.com/neurobagel/recipes/issues/7

- Set fetchOnServer to false
- Updated DOCKERFILE `entrypoint` to use `generate` instead of `build`
- Added trailing slash to the `API_QUERY_URL` in index

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -26,7 +26,7 @@ jobs:
       - name: Run end to end tests
         uses: cypress-io/github-action@v6
         with:
-          build: npm run build
+          build: npm run generate
           start: npm start
           component: false
 
--- a/cypress/e2e/APIRequest.cy.js
+++ b/cypress/e2e/APIRequest.cy.js
@@ -4,11 +4,11 @@ describe('API request', () => {
   it('Intercepts the request sent to the API and asserts over the request url', () => {
     cy.intercept({
       method: 'GET',
-      url: '**/query/?*',
+      url: 'query/?*',
     }, mixedResponse).as('call');
     cy.intercept({
       method: 'GET',
-      url: '**/nodes/',
+      url: '/nodes/',
     }).as('getNodes');
     cy.visit('/?node=OpenNeuro');
     // We need to wait for the fetch to complete and populate the
--- a/cypress/e2e/APIRequest.cy.js
+++ b/cypress/e2e/APIRequest.cy.js
@@ -4,15 +4,18 @@ describe('API request', () => {
   it('Intercepts the request sent to the API and asserts over the request url', () => {
     cy.intercept({
       method: 'GET',
-      url: 'query/?*',
+      url: '**/query/?*',
     }, mixedResponse).as('call');
+    cy.intercept({
+      method: 'GET',
+      url: '**/nodes/',
+    }).as('getNodes');
     cy.visit('/?node=OpenNeuro');
-    cy.get('[data-cy="Min Age-continuous-field-input"]').type('10');
-    cy.get('[data-cy="Max Age-continuous-field-input"]').type('30');
     // We need to wait for the fetch to complete and populate the
     // dropdown with nodes and selecting OpenNeuro before making the request
-    // eslint-disable-next-line
-    cy.wait(1000);
+    cy.wait('@getNodes');
+    cy.get('[data-cy="Min Age-continuous-field-input"]').type('10');
+    cy.get('[data-cy="Max Age-continuous-field-input"]').type('30');
     cy.get('[data-cy="submit-query"]').click();
     cy.wait('@call').its('request.url').should('contains', '&min_age=10&max_age=30');
   });
--- a/cypress/e2e/APIRequest.cy.js
+++ b/cypress/e2e/APIRequest.cy.js
@@ -9,6 +9,10 @@ describe('API request', () => {
     cy.visit('/?node=OpenNeuro');
     cy.get('[data-cy="Min Age-continuous-field-input"]').type('10');
     cy.get('[data-cy="Max Age-continuous-field-input"]').type('30');
+    // We need to wait for the fetch to complete and populate the
+    // dropdown with nodes and selecting OpenNeuro before making the request
+    // eslint-disable-next-line
+    cy.wait(1000);
     cy.get('[data-cy="submit-query"]').click();
     cy.wait('@call').its('request.url').should('contains', '&min_age=10&max_age=30');
   });
--- a/cypress/e2e/Alert.cy.js
+++ b/cypress/e2e/Alert.cy.js
@@ -4,8 +4,15 @@ describe('Alert', () => {
     cy.get('[data-cy="openneuro-alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
     cy.visit('/?node=Parkinson\'s Progression Markers Initiative');
     cy.get('[data-cy="openneuro-alert"]').should('not.exist');
+    // We need to wait for the fetch to complete and populate the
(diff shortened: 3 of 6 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Run end to end tests
         uses: cypress-io/github-action@v6
         with:
-          build: npm run build
+          build: npm run generate
```
surchs, line 29: oh wow, nice catch!

cypress/e2e/APIRequest.cy.js
```diff
@@ -9,6 +9,10 @@ describe('API request', () => {
     cy.visit('/?node=OpenNeuro');
     cy.get('[data-cy="Min Age-continuous-field-input"]').type('10');
     cy.get('[data-cy="Max Age-continuous-field-input"]').type('30');
+    // We need to wait for the fetch to complete and populate the
+    // dropdown with nodes and selecting OpenNeuro before making the request
+    // eslint-disable-next-line
+    cy.wait(1000);
```
surchs: I think it'd be cleaner here to set up another intercept for the fetch call, in the same way that we do it in lines 5-8 for the query. Then we can wait explicitly for that call to finish and don't have to set this specific wait.

cypress/e2e/Alert.cy.js
```diff
@@ -4,8 +4,15 @@ describe('Alert', () => {
     cy.get('[data-cy="openneuro-alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
     cy.visit('/?node=Parkinson\'s Progression Markers Initiative');
     cy.get('[data-cy="openneuro-alert"]').should('not.exist');
+    // We need to wait for the fetch to complete and populate the
+    // dropdown with nodes before searching for OpenNeuro
+    // eslint-disable-next-line
+    cy.wait(1000);
```
surchs: same here, I'd say let's intercept the fetch route and wait for it explicitly!

cypress/e2e/Alert.cy.js
```diff
@@ -4,8 +4,15 @@ describe('Alert', () => {
...
     cy.get('[data-cy="node-field"]').type('OpenNeuro{enter}');
     cy.get('[data-cy="openneuro-alert"]').should('be.visible');
+    // TODO: remove once the alert is fixed
+    // eslint-disable-next-line
+    cy.wait(2000);
```
surchs, line 21: What is this wait for here?

nuxt.config.js
```diff
@@ -19,8 +19,8 @@ export default {
     link: [
       { rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' },
     ],
-    target: 'static',
   },
+  target: 'static',
```
surchs, line 23: :100: :tada: :sparkle:

pages/index.vue
```diff
@@ -105,39 +104,11 @@ export default {
...
-    // for valid query parameters that refer to selected nodes.
-    // If we find any valid query parameters in the URL, then we
-    // update the app with these nodes selected!
+
+    // Since fetchOnServer is set to false the order of execution of
+    // hooks is such that fetch hook is called after mounted and therefore
+    // all logic related to updating the route needs to be in fetch since
+    // nodes are not available in the mounted hook.
     const { node: nodeName } = this.$route.query;
     if (nodeName !== undefined) {
```
surchs, line 114: :+1:

surchs (commented):
Thanks for the PR @rmanaem :tada: and for figuring out the fetch hook situation! Everything looks great functionality wise.

But I'd like to replace the `cy.wait(1000)`... statements in the tests with something more explicit for the events we're waiting for. We're already doing that for other requests, so I think we should be able to just reuse this: https://docs.cypress.io/api/commands/wait#Wait-for-a-specific-request-to-respond

cypress/e2e/Alert.cy.js
rmanaem: The toasts block the `dismiss` button on the alert so we have to wait for them to disappear.

rmanaem: This will be addressed in #328

cypress/e2e/APIRequest.cy.js
```diff
@@ -4,9 +4,16 @@ describe('API request', () => {
   it('Intercepts the request sent to the API and asserts over the request url', () => {
     cy.intercept({
       method: 'GET',
-      url: 'query/?*',
+      url: '**/query/?*',
```
surchs: This is probably ok but why do we need to add the wildcard here?

cypress/e2e/APIRequest.cy.js
```diff
@@ -4,9 +4,16 @@ describe('API request', () => {
...
     }, mixedResponse).as('call');
+    cy.intercept({
+      method: 'GET',
+      url: '**/nodes/',
```
surchs: Similarly here. If this is just in case we'll do something in the future that would need these wildcards then I'd say let's wait until we do

cypress/e2e/Alert.cy.js
```diff
@@ -1,11 +1,24 @@
...
+      url: '**/nodes/',
+    }).as('getNodes');
     cy.visit('/?node=All');
+    // We need to wait for the fetch to complete and populate the
+    // dropdown with nodes before searching for OpenNeuro
+    cy.wait('@getNodes');
     cy.get('[data-cy="openneuro-alert"]').should('be.visible').should('contain', ' The OpenNeuro node is being actively annotated at the participant');
-    cy.visit('/?node=Parkinson\'s Progression Markers Initiative');
+    cy.get('[data-cy="node-field"]').type('Parkinson\'s Progression Markers Initiative{enter}');
+    cy.get('[data-cy="node-field"]').contains('All').find('.vs__deselect').click();
```
surchs, line 13: I'm guessing you need to do this because doing a new visit would also fire Fetch again?

In that case 👍 , but we should probably also have "All" disappear on its own when the user selects a node directly. Mind making an issue for that?

surchs (approved):
Looks good @rmanaem , I left two 🍒  comments but otherwise 🧑‍🍳

cypress/e2e/Alert.cy.js
rmanaem: Done https://github.com/neurobagel/query-tool/issues/344

cypress/e2e/APIRequest.cy.js
rmanaem: Not necessary

cypress/e2e/APIRequest.cy.js
rmanaem: Not necessary, left over from playing around with matching

PR #378, opened 2024-03-01
[FIX] Fixed `apiQueryURL` to always end with a slash

### LINKED ISSUE #22: Ensure apiQueryURL ends in slash in the entire app, not just when sending a query
in neurobagel/query-tool#212 we added an automatic adding of a slash to the query URL:

https://github.com/neurobagel/old-query-tool/blob/81a690fd74e0c08a95079a26622f4d4ed018517a/components/QueryForm.vue#L153-L158

but this only affects the sending of the queries, not the overall apiQueryURL (e.g. for listing the nodes and options):

https://github.com/neurobagel/old-query-tool/blob/81a690fd74e0c08a95079a26622f4d4ed018517a/nuxt.config.js#L55-L58

We should make sure we add a slash if one is missing at the deepest level (maybe even immediately when we read the .env file)

samadpls:
- Closes  https://github.com/neurobagel/react-query-tool/issues/22

- Added optional chaining for API_QUERY_URL in publicRuntimeConfig"

## Checklist

- [X] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see https://neurobagel.org/contributing/pull_requests for more info)_
- [X] PR links to GitHub issue with mention `Closes #XXXX`
- [X] Tests pass
- [X] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -53,7 +53,7 @@ export default {
 
   // Needed for environment variable to be visible in client-side code: https://v2.nuxt.com/tutorials/moving-from-nuxtjs-dotenv-to-runtime-config/#migrating-to-the-nuxt-runtime-config-from-nuxtjsdotenv
   publicRuntimeConfig: {
-    apiQueryURL: process.env.API_QUERY_URL,
+    apiQueryURL: process.env.API_QUERY_URL?.endsWith('/') ? process.env.API_QUERY_URL : `${process.env.API_QUERY_URL}/`,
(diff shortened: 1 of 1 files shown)


surchs:
Thanks @samadpls but we're going to archive this repo in favour of https://github.com/neurobagel/query-tool and the issue was addressed there. So I'll close this PR

